PR #5, opened 2019-01-10
Stream reset

backkem:
This PR implements SCTP Stream reset to enable closing the upper layer data channels.

Skipped:
- [Re-transmission](https://tools.ietf.org/html/rfc6525#section-5.1.1) of the re-config chunks.
- ASCONF, the ASCONF-ACK, and the AUTH chunks. ([rfc5061](https://tools.ietf.org/html/rfc5061#section-4.2.7) & [rfc4895](https://tools.ietf.org/html/rfc4895))
- Strict re-config chunk validation.


backkem:
Maybe this is a good time to add CI as well?

chunk_reconfig_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	for i, tc := range tt {
+		actual := &chunkReconfig{}
+		err := actual.unmarshal(tc.binary)
+		if err != nil {
+			t.Fatalf("failed to unmarshal #%d: %v", i, err)
+		}
+
+		// Reset the raw values
+		/*actual.chunkHeader.raw = []byte{}
```
Sean-Der: Should this stay commented out, or can we delete?

PR #9, opened 2019-01-22
SCTP: Error on short buffer

backkem:
Relates to pions/webrtc#365


Sean-Der:
I always bomb out on read, should we do this for other packages also?

backkem:
Yea, having a uniform way of handling this seems like a good idea.

Issue #11, opened 2018-07-24
Add T3-rtx timer and congestion control


kc5nra:
https://tools.ietf.org/html/rfc4960#section-6.3.3

enobufs:
One consideration:
getPayloadDataToSend() method (will land on master soon) has time.Now() inside it. We should put it outside to make testing easier.

kc5nra:
> One consideration:
> getPayloadDataToSend() method (will land on master soon) has time.Now() inside it. We should put it outside to make testing easier.

As well as time.Since

enobufs:
I am going to start working on this issue form this weekend. The implementation would somewhat relate to congestion control, #14 and likely #21 would be addressed all together.

During the work of pions/webrtc#334, I ran a test with lossy connections using Network Link Conditioner. I noticed nearly 40% of the time, data channel establishment stalled (at 10% packet loss ration on both directions), which is due to the lack of INIT and COOKIE ECHO retransmissions.  This is very crucial for the real world use cases.  So, I am thinking about implementing those in this work as well.

Retransmission timers missing:
* T1-init timer
* T1-cookie timer
* T2-shutdown timer (note1)
* T3-rtx timer
* T5-shutdown-guard timer (note1)
* Heartbeat timer (note2)

> Note 1: this may be done after shutdown is fully implemented in a separat issue

> Note 2: this may be done after heartbeat is implemented in a separat issue

enobufs:
I created a separate issue #24 for T1-init/cookie timers as I'd like to make these timers available earlier.

enobufs:
Most implementation is done.  Now I am writing tests, which may take ~10 more hours. Once the tests are done, I will push my code for review.
Here's [my implementation note](https://docs.google.com/document/d/18HxKodWkPPfZpKTddQQvepux5MOS3QWMfK1eJYXQdz4/edit#)

enobufs:
I have just realized that the `bufferredAmout` and its threshold are per-stream (per-datachannel) properties. I will need to update the code...

enobufs:
We need to determine the size of the receive window size (rwnd), which is currently set to 64KB.
Initial ssthresh value is set to the rwnd as well (as recommended in RFC 4960)

[...]

PR #13, opened 2019-01-28
[WIP] Partial Reliability

backkem:
This PR aim to add support for [Partial Reliability](https://tools.ietf.org/html/rfc3758).

cc @enobufs

Relates to pions/webrtc#334


enobufs:
@backkem I merged my branch into pr-sctp branch (I did like my branch name..) as you already had PR for it.

enobufs:
Here's my [implementation note](https://docs.google.com/document/d/1i4FgCd0gZc1aHh7zcA_m81yIzvsxHXejTMKpcu0SM4s) - just in case that helps for your review.

backkem:
Fixed the commit lint issues. Will try to get up-to-speed with the changes tomorrow.

enobufs:
Great! Thank you.  I also need to remove prinfs.  I wish we had a logger for sctp module, maybe we can introduce in the next round.

association_test.go
```diff
@@ -224,3 +225,1031 @@ func (c *dumbConn) SetReadDeadline(t time.Time) error {
 func (c *dumbConn) SetWriteDeadline(t time.Time) error {
 	return nil
 }
+
+////////////////////////////////////////////////////////////////////////////////
+// testConn, connBridge for emulating dtls.Conn
+
+type testConn struct {
```
backkem: Maybe we should conciser moving this to https://github.com/pions/transport ?

association.go
```diff
@@ -599,6 +683,82 @@ func (a *Association) handleReconfig(c *chunkReconfig) ([]*packet, error) {
...
+		a.payloadQueue.pop(a.peerLastTSN + 1) // may not exist
+		a.peerLastTSN++
+	}
+
+	// From RFC 3758 Sec 3.6:
+	//   .. and then MUST further advance its cumulative TSN point locally
+	//   if possible
+	// Meaning, if peerLastTSN+1 points to a chunk that is received,
+	// advance peerLastTSN until peerLastTSN+1 points to unreceived chunk.
+	_, popOk := a.payloadQueue.pop(a.peerLastTSN + 1) // may not exist
```
kc5nra: Could we combine these into one for? Incoming pseudo-code
```
for _, popOk = a.payloadQueue.pop(a.peerLastTSN + 1) {
	a.peerLastTSN++
}

File: association_test.go [hunk 3]
```diff
@@ -224,3 +225,1031 @@ func (c *dumbConn) SetReadDeadline(t time.Time) error {
...
+
+		fwdtsn := a.createForwardTSN()
+
+		assert.Equal(t, uint32(10), fwdtsn.newCumulativeTSN, "should be able to serialize")
+		assert.Equal(t, 1, len(fwdtsn.streams), "there should be one stream")
+		assert.Equal(t, uint16(1), fwdtsn.streams[0].identifier, "si should be 1")
+		assert.Equal(t, uint16(2), fwdtsn.streams[0].sequence, "ssn should be 2")
+	})
+
+	t.Run("forward two abndoned with the same SI", func(t *testing.T) {
```
kc5nra: teensy spelling abandoned

chunktype.go
```diff
@@ -24,6 +24,7 @@ const (
 	CWR              chunkType = 13
 	SHUTDOWNCOMPLETE chunkType = 14
 	RECONFIG         chunkType = 130
+	FORWARDTSN       chunkType = 192
```
kc5nra: Should we take a chance to rename these? All caps is both against _most_ of golangs style (besides posix things, etc.) and also give it unnecessary global scope

reassembly_queue.go
```diff
@@ -1,116 +1,329 @@
...
+	return chunkSet
 }
 
-type reassemblyQueue struct {
-	messageQueue     payloadDataMessageArray
-	unorderedMessage payloadDataMessage
-	expectedSeqNum   uint16
+func (r *reassemblyQueue) isReadable() bool {
+	// Check unordered first
+	if len(r.unordered) > 0 {
```
kc5nra: Does it make sense the mark either of these as readable when packets are added to fast-exit?

reassembly_queue.go
```diff
@@ -1,116 +1,329 @@
...
 
-func (r *reassemblyQueue) push(p *chunkPayloadData) {
-	if p.unordered {
-		r.unorderedMessage.fragmentQueue = append(r.unorderedMessage.fragmentQueue, p)
-		r.unorderedMessage.length += len(p.userData)
-		r.unorderedMessage.payloadType = p.payloadType
-		return
+func (r *reassemblyQueue) read(buf []byte) (int, PayloadProtocolIdentifier, error) {
+	var cset *chunkSet
+	// Check unordered first
```
kc5nra: Is there a rule to state which is checked first?

stream.go
```diff
@@ -127,23 +176,38 @@ func (s *Stream) packetize(raw []byte, ppi PayloadProtocolIdentifier) []*chunkPa
...
+	//   ordered delivery and reliable transmission.
+	unordered := ppi != PayloadTypeWebRTCDCEP && s.unordered
+
 	var chunks []*chunkPayloadData
 	for remaining != 0 {
 		l := min(s.association.myMaxMTU, remaining)
-		chunks = append(chunks, &chunkPayloadData{
+		chunk := &chunkPayloadData{
 			streamIdentifier:     s.streamIdentifier,
 			userData:             raw[i : i+l],
```
kc5nra: Let's change `l` to something else -- I had to look several times before I realized it wasn't a 1

association_test.go
enobufs: Agreed.

association.go
```diff
@@ -431,27 +439,31 @@ func (a *Association) handleInitAck(p *packet, i *chunkInitAck) (*packet, error)
 
 // The caller should hold the lock.
 func (a *Association) handleData(d *chunkPayloadData) []*packet {
-	reply := make([]*packet, 0)
-
-	a.payloadQueue.push(d, a.peerLastTSN)
+	added := a.payloadQueue.push(d, a.peerLastTSN)
+	if added {
+		// Pass the new chunk to session level as soon as it arrives
```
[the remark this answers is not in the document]
enobufs: Could you please elaborate on "implementing the sna32 logic as a type."?

association.go
enobufs: Agreed. Will do.

association_test.go
enobufs: oops - will fix

reassembly_queue.go
enobufs: I was debating on that... If we introduce an internal flag, "readable", then it has to be maintained by many places, where the current one simplifies it with a bit of cost of evaluating the arrays...

reassembly_queue.go
enobufs: I have not come across any spec on that but I just followed how it was implemented (was checking the unordered packet first), and also with my assumption (for 2c), an application would choose to use "unordered" delivery mostly because for smaller latency. I had this impression from the text in RFC 4960 sec 6.6:
```
   IMPLEMENTATION NOTE: When sending an unordered DATA chunk, an
   implementation may choose to place the DATA chunk in an outbound
   packet that is at the head of the outbound transmission queue if
   possible.
```

stream.go
enobufs: Ok, will do!

association.go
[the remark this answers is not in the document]
backkem: Sorry, let me give an example:
```go
type serialNumber uint32

func (i1 serialNumber) LT(i2 serialNumber) bool {
	return (i1 < i2 && i2-i1 < 1<<31) || (i1 > i2 && i1-i2 > 1<<31)
}
```
There may be more opinions on this but I think it makes it more obvious that they should be handled in this way. As I said, feel free to ignore if you disagree.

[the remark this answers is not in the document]
enobufs: I guess, something like this:
```
type SNA32 struct uint32
func (sna SNA32) GreaterThan(val SNA32) bool {...}
   :
```
I think it's a good idea.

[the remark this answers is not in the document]
backkem: Indeed, I wouldn't export the type if it isn't needed thought.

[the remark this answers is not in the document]
enobufs: @backkem As I was making changes for this idea, I realized I have to change many files that have sequence numbers, all chunks that have tsn, ssn, for instance - too many changes for a little gain.  Let me ask you again if we want to do this.  Using sna32 type does not prevent us from using standard inequality comparison either. (unless we ues a comple type than primitive type.)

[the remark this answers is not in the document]
enobufs: Copied the work to [my gist](https://gist.github.com/enobufs/3ef71a587ba2b9ece55dd54182b8d48f) for future reference.

association_test.go
enobufs: done @f8bfd

association.go
enobufs: Fixed @98bccc

association_test.go
enobufs: done

association.go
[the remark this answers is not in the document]
backkem: I say leave it for now.

chunktype.go
enobufs: Fixed at e928d.  I prefixed with 'ct' (chunk type).  E.g. ctForwardTSN

stream.go
enobufs: Fixed at e928d.  `l` => `fragmentSize`

enobufs:
@backkem @kc5nra @Sean-Der I've made changes per your feedback.  Those that I wasn't sure left open with my comments. Any thoughts? (I will work on datachannel tomorrow)

association.go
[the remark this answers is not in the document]
enobufs: (will fix the typo in the next commit: session => stream)

reassembly_queue.go
kc5nra: I'm totally fine with that answer! Caching things should be a post-evaluation, hotspot squashing activity

reassembly_queue.go
kc5nra: Sounds good! I wasn't sure if it was noted in the RFC specifically (and if so, best to leave a comment reflecting that)

association.go
```diff
@@ -697,6 +855,56 @@ func (a *Association) sendPayloadData(chunks []*chunkPayloadData) error {
 	return nil
 }
 
+// getPayloadDataToSend updates chunk status and returns a list of packets we can send.
+// The caller should hold the lock.
+func (a *Association) getPayloadDataToSend(onlyUnsent bool) []*packet {
```
kc5nra: As I was thinking.. does it make sense to pass in the time into this so that it can be tested? Rather than relying on "Now", which makes precise testing difficult

enobufs: True. As we need to make changes around this method when we implement T3-rtx timer, let's apply that idea at that time. I will leave a note on #11.

reassembly_queue.go
enobufs: Leave a note is a good idea! Will do it.

PR #22, opened 2019-03-05
[WIP] Implement retransmission timers (T1-init/cookie)

enobufs:
This is just the beginning of the retransmission timer implementation

Relates to #11


enobufs:
Now this PR is for #24 (split from the original #11).

go.mod
```diff
@@ -1,7 +1,8 @@
 module github.com/pions/sctp
 
 require (
-	github.com/pions/transport v0.2.0
+	github.com/davecgh/go-spew v1.1.1 // indirect
```
backkem: Maybe try removing this line followed by `go mod tidy`. I've noticed these sneaking in but I don't think they are really necessary.

enobufs: Thank you! I will do it once pions/transport#9 (this PR depends on) has merged into master.

PR #38, opened 2019-04-09
Calculate rwnd using reassembly queues

### LINKED ISSUE #37: 'rwnd' never becomes 0
### Your environment.
  - Version: v2.0.3 or earlier
  - Browser: n/a

### What did you do?
Sent 100 MB of data over a data channel, using MaxRetransmits set to 0 or 1, then the transmission stops due to an issue reported as pions/ice#12.

### What did you expect?
All 100 MB of data transfer should complete successfully.

### What happened?
The issue pions/ice#12 is a separate problem, but obviously, SCTP is pushing more than 1MB of data even though the receiver window is set to 128KB.

"1MB" is the maximum number of bytes ICE would buffer at most.

I confirmed that when this happens, the receiver is still reporting a_rwnd (advertised receiver window size) is 128KB (=fully available).

I noticed that the current receiver window size is calculated only by the payloadQueue size. This is because received DATA chunks are almost immediately handed off to the stream layer, and DATA chunks in the payloadQueue will be removed as soon as a.peerLastTSN advances, while those DATA chunks are still in the reasemblyQueue and waiting to be read by the application.

The rwnd to be advertised to the sender should use the number of bytes stored in the reassemblyQueue in its calculation, so that sender will stop sending data beyond the amount of data, or rwnd, the application has not read.

enobufs:
#### Description
See #37 for details.
Now it generates correct rwnd and reports it to the sender.

#### Reference issue
Resolves #37

#### Notes
This did not resolve the "slow reader" case mentioned in #37. What's filling the packetio/buffer was actually the control chunks (Forward TSN), not the DATA chunk. However, counting rwnd using reassembly queues is the right thing to do.

### DIFF
--- a/association.go
+++ b/association.go
@@ -644,6 +644,8 @@ func (a *Association) handleData(d *chunkPayloadData) []*packet {
 			// Pass the new chunk to stream level as soon as it arrives
 			a.payloadQueue.push(d, a.peerLastTSN)
 			s.handleData(d)
+		} else {
+			a.log.Debugf("receive buffer full. dropping DATA with tsn=%d", d.tsn)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


enobufs:
Changes in this branch are part of PR #42 which has landed on the master.

PR #39, opened 2019-04-13
[WIP] Implemented delayed ack

### LINKED ISSUE #32: Delayed Ack
## Summary
I'd consider this as a post-v2 work item (no API change required)
The support of congestion control addressed in #11 has drastically improved the performance, but it is not optimal yet. To improve it further, we will need to implement delayed ack to reduce the number of packets per sec. 

## Motivation
Say if you are handling 20MB/s of traffic, as the current MTU size (for SCTP) is set to 1200 bytes, SCTP (sender) is handling 17500 packets of outgoing DATA chunks and the same number of SACK packets incoming per second. That's a lot of CPU usage.  We can drastically reduce the number of SACK chunks by implementing the delayed ack. (the `handleSelectiveAck` routine is the most complex/expensive routine in sctp, and we could reduce 17500 SACK chunks to just 5! - in 20Mbps/200msec delayed ack). We should also piggy-back the ack on the outgoing DATA chunks also as recommended by RFC 4960.

## Describe alternatives you've considered
No alternative I can think of right now.

## TODO
* Profile CPU usage first
* Implement delayed Ack (up to 200 msec - check the spec)
* Implement piggy-back

enobufs:
#### Description
* Calculate rwnd using reassembly queues (#37)
* Ack timer for delayed ack
* Use SACK-IMMEDIATELY ([RFC 7053](https://tools.ietf.org/html/rfc7053))
* Send SACK when packet loss occurs (bug fix)

#### Reference issue
Resolves #32 and #37

#### Notes
* TODO
    - performance tuning with I-bit flag
    - test with datachannel and webrtc packages
    - benchmark


backkem:
It seems like there is still some slow receiver issue going on in the `libp2p/go-libp2p-webrtc-direct` tests: https://travis-ci.org/libp2p/go-libp2p-webrtc-direct/builds/520214803.
On my local machine it isn't nearly as bad but I'm also seeing `popPendingDataChunksToSend` reaching ~1000 chunks in one of the tests.

enobufs:
@backkem There are many "packetloss !!!!!!!" in your CI log.  I wonder how did the loss occur.  When loss is detected in handleData() routine, it tries to send SACK every time (as RFC says)...

backkem:
Maybe it's due to the buffering on ICE level? It's all over loopback so it must almost be due to a slow reader.

enobufs:
Possible. I saw ICE buffer becoming full (pions/ice#12) earlier (before delayed-ack). What filling the buffer was a bunch of SCTP control packets, and SCTP's flow control cares only about DATA chunks with TSN...  Looks like we have to find out why it is reading slow, first...
Is the failing test using partial reliability channel?

Sean-Der:
Hey @enobufs 

I am working on some ICE issues that were brought up (so no rush) but what do you think of hitting HN/reddit with v2.0.0 after this PR lands? 

I want us to put out something we are proud of, and I think this is such a great improvement it is worth waiting. Also anything I can do (or the greater team) to help with this? I don't want to put all the work on you.

thanks again for all the hard work :)

enobufs:
> what do you think of hitting HN/reddit with v2.0.0 after this PR lands?
@Sean-Der Sounds good to me!

The current sctp@v1.5.0 is definitely unacceptable... This PR already make a big improvement, but I am not confident if this makes everyone happier yet. I am trying to test over a bad connection also. Are you guys able to wait till this weekend? @backkem how do I run the test 'SubtestStress1Conn100Stream100Msg'? (there was a build error on my mac...)

Sean-Der:
Absolutely! No rush at all @enobufs :)

This is all in our spare time, so I don't want to make it stressful/rush anything out! I just want to make something we are all proud of, and make sure people stay contributors for a long time.

backkem:
@enobufs I'm using https://github.com/libp2p/go-libp2p-webrtc-direct/pull/29 plus replace statements in the go.mod file so i can mess with sctp.

enobufs:
@backkem the tests (SubtestStress1Conn100Stream100Msg and SubtestStress1Conn100Stream100Msg10MB) pass on my mac. I see stats from sctp DEBUG log:
```
sctp DEBUG: 15:01:42.515864 association.go:349: [0xc000410000] stats nDATAs (in) : 20101
sctp DEBUG: 15:01:42.515873 association.go:350: [0xc000410000] stats nSACKs (in) : 399
sctp DEBUG: 15:01:42.515881 association.go:351: [0xc000410000] stats nT3Timeouts : 0
sctp DEBUG: 15:01:42.515889 association.go:352: [0xc000410000] stats nAckTimeouts: 8
sctp DEBUG: 15:01:42.515899 association.go:353: [0xc000410000] stats nFastRetrans: 0
```
The delayed-ack seems to be working well (nSACKs << nDATA). I saw one fast retranmission, but it recovered right away (fast recovery). nAckTimeouts is a little more than I'd expect but it could happen if DATA chunks arrives with a large (> 200msec) delay.  I will check how much the ICE buffer is queueing next.  Tranvis-ci's cpu might be very constraint? (if incoming data is queued at lower layer, sctp can not do much to it...)

enobufs:
@hugo pointed out a good one. The stream layer has one lock over which ReadSCTP and WriteSCTP are competing. I am thinking about localizing (separating) the read side locking inside the reassembly queue so that the access to the reassembly queue wouldn't block or be blocked by other routines.

enobufs:
[syscall.pdf](https://github.com/pion/sctp/files/3087622/syscall.pdf)
Looking at the trace output (go tool trace), it appears, what association's readLoop is blocked by is not the mutex of stream layer, but the syscall "sendto". (underlying (UDP) socket is a blocking socket)
To solve the "slow-reader" problem, we'd probably need a drastic change in SCTP, such as stop sending data (a reply) on handleInbound(), and use another goroutine for the immediate replies (control packets), which would come with another issue - how much can sctp layer should buffer those immediate reply?
We should separate this issue from this one (delayed-ack), give more thoughts/analysis on it.

enobufs:
@backkem, I found a bug in fast retransmission, not recovering well from contiguous losses. (just wanted to let you know I am working on it)

enobufs:
@backkem  accessing a.send() from **multiple goroutines** was a mistake, causing massive packet reorder which triggers the bug I mentioned.  I am going to introduce writeLoop to solve #40 altogether.

enobufs:
This branch is part of PR #41 which has already landed on master.

PR #41, opened 2019-04-23
[WIP] Introduce writeLoop

### LINKED ISSUE #40: Slow-reader problem
## The issue
It appears that SCTP is reading incoming data slow in a high bandwidth usage, causing a buffer (packetio) at ICE layer being filled up with a lot of incoming data (cause a lot of latency before reaching SCTP layer) - pion/ice#12.

(Excerpt from the discussion in the PR #39)
[syscall.pdf](https://github.com/pion/sctp/files/3087622/syscall.pdf)
Looking at the trace output (go tool trace), it appears, what association's readLoop is blocked by is not the mutex of stream layer, but the syscall "sendto". (underlying (UDP) socket is a blocking socket)
To solve the "slow-reader" problem, we'd probably need a drastic change in SCTP, such as stop sending data (a reply) on handleInbound(), and use another goroutine for the immediate replies (control packets), which would raise another issue - how much can sctp layer buffer those immediate replies?

Related issue: pion/ice#12 and #32

## Approach (proposal)
* Introduce another goroutine, `writeLoop` in association layer
* WriteSCTP() stores data to pending data, wake up the `writeLoop` and return immediately.
* handleInbound() (the readLoop routine) DO NOT CALL association.send(). Instead, it should schedule the transmission of immediate control packets for `writeLoop` to send it, and exit the routine as soon as possible (similar to delayed-ack, but immediately)
* Also consider avoiding read and write routines waiting on the same mutex at the stream layer. (introduce a mutex inside reassembly queue, etc.

TODO: evaluate if the above proposal is viable first.

enobufs:
#### Description
To solve slow-reader problem, writeLoop (goroutine) has been introduced.
This writeLoop routine is responsible for calling a.send() for all outgoing packets.

#### Reference issue
Fixes #40


enobufs:
@backkem can you try this branch with your libp2p test on CI?

Sean-Der:
Fantastic work @enobufs 

Anything I can do just tell me @backkem @albrow! 

getting ipfs unblocked (then announcing v2.0.0) is going to be huge. Would love to have more active users! been getting a lot of issues/asks all the sudden but not many more contributors

ack_timer.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ackTimerObserver is the inteface to an ack timer observer.
+type ackTimerObserver interface {
+	onAckTimeout()
+}
+
+// ackTimer provides the retnransmission timer conforms with RFC 4960 Sec 6.3.1
+type ackTimer struct {
+	observer ackTimerObserver
+	interval float64
```
hugoArregui: maybe this can be a time.Duration instead?

association.go
```diff
@@ -384,18 +450,118 @@ func (a *Association) handleInbound(raw []byte) error {
...
+
+	packets := []*packet{}
+	if a.controlQueue.size() > 0 {
+		packets = append(packets, a.controlQueue.popAll()...)
+	}
+
+	state := a.getState()
+
+	if state == established {
+		if !a.willRetransmitDataChunks {
```
hugoArregui: tiny suggestion: IMHO I think inverting the conditions to have the positive case first will make it more readable

association.go
```diff
@@ -384,18 +450,118 @@ func (a *Association) handleInbound(raw []byte) error {
...
+				//  3)  Determine how many of the earliest (i.e., lowest TSN) DATA chunks
+				//      marked for retransmission will fit into a single packet, subject
+				//      to constraint of the path MTU of the destination transport
+				//      address to which the packet is being sent.  Call this value K.
+				//      Retransmit those K DATA chunks in a single packet.  When a Fast
+				//      Retransmit is being performed, the sender SHOULD ignore the value
+				//      of cwnd and SHOULD NOT delay retransmission for this single
+				//		packet.
+				dataChunkSize := dataChunkHeaderSize + uint32(len(c.userData))
+				if a.mtu-fastRetransSize >= dataChunkSize {
```
hugoArregui: not sure to understand the spec so not sure if this is ok or not, but here you check if the current chunk fits the mtu.. is that right or you meant something along the lines of `a.mtu-fastRetransSize >= (dataChunkSize + fastRetransSize)`?

enobufs: I guess you meant:
`a.mtu >= (dataChunkSize + fastRetransSize)`
Do you think this is more readable?

hugoArregui: I'm so stupid, sorry, disregard my comment, I read `a.mtu-fastRetransSize` as a var :man_facepalming: sorry for the noise

ack_timer.go
enobufs: Agreed.

association.go
enobufs: OK.

ack_timer.go
enobufs: Done

association.go
enobufs: Done

enobufs: Thanks, @hugo for spending time to review the code. I really appreciate it.

Sean-Der:
Hey @backkem would you mind checking out this PR? I would love to know when you feel good about it so we can release v2.0.0 I have pushed it back about a month at this point. 

Or we can decouple the release from these changes.

thanks

enobufs:
I really hope that this could land in v2.0.0 as soon as possible because the performance with the current one on master is pretty bad. (storm of control packets with slow reader causing a huge backlog in ice buffer)... my heart racing...

PR #47, opened 2019-05-07
Export association is closed error

trivigy:
The goal is to ignore the error in case of `pc.Close()`.


enobufs:
I will create another PR.

PR #51, opened 2019-06-19
Patch to properly return an io.ErrShortBuffer

### LINKED ISSUE #50: Stream doesn't check for io.ErrShortBuffer returns from reassemblyQueue.read
### Your environment.
  - Version: Release
  - Browser: N/A
  - Other Information - See https://trac.torproject.org/projects/tor/ticket/28942#comment:14

### What did you do?
Used the pion/webrtc library for the the proxy side of our snowflake anti-censorship system that reads from a WebRTC connection to the client. I used keroserene/go-webrtc as the library on the client side and found data being dropped despite the usage of a reliable channel.

### What did you expect?
I expected to receive all of the data that I sent

### What happened?
Large chunks of data were missing.

### More details:
It turns out Stream is not returning `io.ErrShortBuffer` errors from reassemblyQueue.read. Instead the error is being overwritten and the data is lost. Stream.ReadSCTP should instead return the error.

cohosh:
If the reassemblyQueue returns an io.ErrShortBuffer, this error is
returned instead of being overwritten and lost

#### Description

Since the datachannel read loop https://github.com/pion/webrtc/blob/master/datachannel.go#L292 is checking for io.ErrShortBuffer errors, the best thing to do is simply return this error from Stream.ReadSCTP
 and let the calling functions handle it.

See https://trac.torproject.org/projects/tor/ticket/28942#comment:15

#### Reference issue
Fixes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Michael MacDonald](https://github.com/mjmac) - *Fix races*
 * [Yutaka Takeda](https://github.com/enobufs) - *PR-SCTP, Retransmissions, Congestion Control*
 * [Antoine Baché](https://github.com/Antonito) - *SCTP Profiling*
+* [Cecylia Bocovich](https://github.com/cohosh) - *Fix SCTP reads*
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/association_test.go
+++ b/association_test.go
@@ -696,6 +696,42 @@ func TestAssocReliable(t *testing.T) {
 		assert.False(t, s0.reassemblyQueue.isReadable(), "should no longer be readable")
 		closeAssociationPair(br, a0, a1)
 	})
+
+	t.Run("short buffer", func(t *testing.T) {
+		const si uint16 = 1
+		const msg = "Hello"
+		br := test.NewBridge()
+
+		a0, a1, err := createNewAssociationPair(br, ackModeNoDelay)
+		if !assert.Nil(t, err, "failed to create associations") {
+			assert.FailNow(t, "failed due to earlier error")
+		}
+
+		s0, s1, err := establishSessionPair(br, a0, a1, si)
+		assert.Nil(t, err, "failed to establish session pair")
+
+		assert.Equal(t, 0, a0.bufferedAmount(), "incorrect bufferedAmount")
+
+		n, err := s0.WriteSCTP([]byte(msg), PayloadTypeWebRTCBinary)
+		if err != nil {
+			assert.FailNow(t, "failed due to earlier error")
+		}
+		assert.Equal(t, len(msg), n, "unexpected length of received data")
+		assert.Equal(t, len(msg), a0.bufferedAmount(), "incorrect bufferedAmount")
+
+		br.Process()
+
+		buf := make([]byte, 3)
+		n, ppi, err := s1.ReadSCTP(buf)
+		assert.Equal(t, err, io.ErrShortBuffer, "expected error to be io.ErrShortBuffer")
+		assert.Equal(t, n, 0, "unexpected length of received data")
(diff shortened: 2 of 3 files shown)


cohosh:
Ah, I'm seeing an issue now about how the caller should handle the error. I think it's a good idea to allow the caller to try again with a larger buffer. Right now the reassemblyQueue increments the SSN and removes the chunkset from the list. I think the reassembly queue should leave the chunkset in the list until the read returns without error.

enobufs:
TODO: investigate WebRTC protocol spec, and EOR (e.g. [this article](https://lgrahl.de/articles/demystifying-webrtc-dc-size-limit.html)

Sean-Der:
This is super exciting, thanks for the contribution @cohosh :)

I added you to the Pion organization, when the PR gets approved you will be able to merge it right away. I really hope we can make `pion/webrtc` work for snowflake, I would be very proud to think we helped make Tor better :) also I think moving away from Google controlled software to something owned by the community is better for everyone.

If there is anything I can do personally I am always happy to help! We have a [Slack](https://pion.ly/slack) and always feel free to drop me an email sean @ pion.ly :)

cohosh:
Thanks! It's probably a good idea to address the changes to the datachannel read loop along with this change: https://github.com/pion/webrtc/pull/719
Otherwise, this change will cause an infinite loop by the caller.

cohosh:
Also you can follow the integration to Snowflake here: https://trac.torproject.org/projects/tor/ticket/28942
Thanks for the interest, we're excited about getting this into a usable and easily buildable state :)

backkem:
@cohosh An alternative to this can be our detach API:
https://github.com/pion/webrtc/blob/master/examples/data-channels-detach/main.go
This allows you to detach the underlying datachannel implementation, giving you direct access to the io.ReadWriter we use internally from which you can read with your own buffer of any size.
As an example, I've used this in [libp2p/go-libp2p-webrtc-direct](https://github.com/libp2p/go-libp2p-webrtc-direct):
https://github.com/libp2p/go-libp2p-webrtc-direct/blob/b213ddc5d17985af4a77a60757cdc5956d71e6b7/transport.go#L26-L30
This enables you to read from the data channel in an idiomatic way:
https://github.com/libp2p/go-libp2p-webrtc-direct/blob/b213ddc5d17985af4a77a60757cdc5956d71e6b7/stream.go#L20

cohosh:
> @cohosh An alternative to this can be our detach API:
> https://github.com/pion/webrtc/blob/master/examples/data-channels-detach/main.go
> This allows you to detach the underlying datachannel implementation, giving you direct access to the io.ReadWriter we use internally from which you can read with your own buffer of any size.
> As an example, I've used this in [libp2p/go-libp2p-webrtc-direct](https://github.com/libp2p/go-libp2p-webrtc-direct):
> 
> https://github.com/libp2p/go-libp2p-webrtc-direct/blob/b213ddc5d17985af4a77a60757cdc5956d71e6b7/transport.go#L26-L30
> 
> This enables you to read from the data channel in an idiomatic way:
> 
> https://github.com/libp2p/go-libp2p-webrtc-direct/blob/b213ddc5d17985af4a77a60757cdc5956d71e6b7/stream.go#L20-L21

Hi @backkem !
Thanks for the response. I took a look at the detached channels, and I think we'd still prefer to be able to use the OnMessage callback instead of implementing our own readLoop (especially since these issues seem to be actual bugs that, if fixed, would solve the issues we're having).

cohosh:
Oh, also just to confirm... it seems detaching the channel still requires this SCTP patch otherwise the io.ErrShortBuffer won't be returned. So channel detachment is really in case you decide to handle the io.ErrShortBuffer differently than this PR: https://github.com/pion/webrtc/pull/719. It would also allow us to specify a larger buffer. Is that a good summary?

cohosh:
Hey, as mention in https://github.com/pion/webrtc/issues/718, I updated this patch to remove the modifications to the reassembly_queue, but keep the `io.ErrShortBuffer` return.

cohosh:
Thanks!

PR #56, opened 2019-08-17
Fix routine leak

at-wat:
#### Description

writeLoop routine can be leaked if the connection is lost when there is no writing packet.
So, writeLoop should also be stopped if readLoop is stopped.
And, closeWriteLoopCh should be closed, even if netConn.Close() returned error.

#### Additional resources

After running the system using webrtc.PeerConnection for a while on the real environment, pprof could find routine leaks.
```
goroutine profile: total 92
79 @ 0x42dcbb 0x43da76 0x78de46 0x45b161
#	0x78de45	github.com/pion/sctp.(*Association).writeLoop+0x1b5	/go/pkg/mod/github.com/pion/sctp@v1.6.4/association.go:426

2 @ 0x42dcbb 0x42dd63 0x404e8d 0x404c65 0x79b63b 0x7ab3f6 0x45b161
#	0x79b63a	github.com/pion/sctp.(*Association).onRetransmissionFailure+0x41a	/go/pkg/mod/github.com/pion/sctp@v1.6.4/association.go:2003
#	0x7ab3f5	github.com/pion/sctp.(*rtxTimer).start.func1+0x175			/go/pkg/mod/github.com/pion/sctp@v1.6.4/rtx_timer.go:158
```
The number of the writeLoop routine seems increased after `retransmission failure: T1-init` on my environment.

Heap memory was also leaked like:
image
and these memory leaks disappeared after applying this PR.

### DIFF
--- a/association.go
+++ b/association.go
@@ -377,6 +377,10 @@ func (a *Association) closeAllTimers() {
 func (a *Association) readLoop() {
 	var closeErr error
 	defer func() {
+		// also stop writeLoop, otherwise writeLoop can be leaked
+		// if connection is lost when there is no writing packet.
+		a.closeWriteLoopOnce.Do(func() { close(a.closeWriteLoopCh) })
+
 		a.lock.Lock()
 		for _, s := range a.streams {
 			a.unregisterStream(s, closeErr)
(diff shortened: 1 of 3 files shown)


association.go
```diff
@@ -406,6 +405,10 @@ func (a *Association) readLoop() {
 	}
 
 	a.log.Debugf("[%s] readLoop exited", a.name)
+
+	// also stop writeLoop, otherwise writeLoop can be leaked
+	// if connection is lost when there is no writing packet.
+	a.closeWriteLoopOnce.Do(func() { close(a.closeWriteLoopCh) })
```
enobufs: @at-wat Here's my small request...  Can you move this line into the defer above, where maybe below the `close(a.readLoopCloseCh)`?  (just to put similar lines together)

at-wat: Sure, it sounds better.

at-wat:
@enobufs I have addressed your comment. PTAL

enobufs:
Thanks @at-wat for your contribution!

PR #58, opened 2019-08-23
Align for 32 bits

hugoArregui:
align for 32bits


enobufs:
@Sean-Der @hugoArregui This is a great addition to the CI test!

PR #73, opened 2019-09-30
Fix unordered message corruption

### LINKED ISSUE #71: Reassembled unordered message is received broken
### Your environment.
  - Version: v1.6.11

### What did you do?
I was writing test with sctp stream configured for unordered delivery.

### What did you expect?
Sending messages, although the order could be different, each messages are identical to the original messages. 

### What happened?
Each message is broken when the message is larger than the max segment size, 1200.

I know the cause.

The assumption of the design was wrong. When I introduced pendingQueue which has ordered and unordered queues assuming that chunks in the unordered queue can be completely unordered. But 'unordered' should only apply to user messages. Not at the chunk level.

When a message is large, fragmented into multiple chunks with the current code, the message could be corrupted. (this has been repro'ed in my local test.  PR is incoming)

enobufs:
Resolves #71

* Added a comparison between received data an original data transmitted to vnet_test.go
* Unordered message was found broken
* Potential reorder in the pendingQueue was not happening (which was fixed and unit-tested)
* found **reassmblyQueue had a bug in handling unordered chunk slices**
* fix was confirmed by the test

### DIFF
--- a/pending_queue.go
+++ b/pending_queue.go
@@ -43,9 +43,11 @@ func (q *pendingBaseQueue) size() int {
 // pendingQueue
 
 type pendingQueue struct {
-	unorderedQueue *pendingBaseQueue
-	orderedQueue   *pendingBaseQueue
-	nBytes         int
+	unorderedQueue      *pendingBaseQueue
+	orderedQueue        *pendingBaseQueue
+	nBytes              int
+	selected            bool
+	unorderedIsSelected bool
 }
 
 func newPendingQueue() *pendingQueue {
@@ -65,6 +67,13 @@ func (q *pendingQueue) push(c *chunkPayloadData) {
 }
 
 func (q *pendingQueue) peek() *chunkPayloadData {
+	if q.selected {
+		if q.unorderedIsSelected {
+			return q.unorderedQueue.get(0)
+		}
+		return q.orderedQueue.get(0)
+	}
+
 	if c := q.unorderedQueue.get(0); c != nil {
 		return c
 	}
@@ -72,15 +81,44 @@ func (q *pendingQueue) peek() *chunkPayloadData {
 }
 
 func (q *pendingQueue) pop(c *chunkPayloadData) error {
-	if c.unordered {
-		popped := q.unorderedQueue.pop()
-		if popped != c {
-			return fmt.Errorf("unexped chunk popped (unordered)")
+	if q.selected {
+		var popped *chunkPayloadData
+		if q.unorderedIsSelected {
+			popped = q.unorderedQueue.pop()
+			if popped != c {
+				return fmt.Errorf("unexped chunk popped (unordered)")
+			}
+		} else {
+			popped = q.orderedQueue.pop()
+			if popped != c {
+				return fmt.Errorf("unexped chunk popped (ordered)")
+			}
+		}
+		if popped.endingFragment {
+			q.selected = false
 		}
 	} else {
-		popped := q.orderedQueue.pop()
-		if popped != c {
-			return fmt.Errorf("unexped chunk popped (ordered)")
+		if !c.beginningFragment {
+			return fmt.Errorf("unexpected q state (should've been selected)")
+		}
+		if c.unordered {
+			popped := q.unorderedQueue.pop()
+			if popped != c {
+				return fmt.Errorf("unexped chunk popped (unordered)")
+			}
+			if !popped.endingFragment {
+				q.selected = true
+				q.unorderedIsSelected = true
+			}
+		} else {
+			popped := q.orderedQueue.pop()
+			if popped != c {
+				return fmt.Errorf("unexped chunk popped (ordered)")
+			}
+			if !popped.endingFragment {
+				q.selected = true
+				q.unorderedIsSelected = false
+			}
 		}
 	}
 	q.nBytes -= len(c.userData)
--- a/reassembly_queue.go
+++ b/reassembly_queue.go
@@ -216,15 +216,15 @@ func (r *reassemblyQueue) findCompleteUnorderedChunkSet() *chunkSet {
 	}
 
 	// Extract the range of chunks
-	chunks := r.unorderedChunks[startIdx : startIdx+nChunks]
+	var chunks []*chunkPayloadData
+	chunks = append(chunks, r.unorderedChunks[startIdx:startIdx+nChunks]...)
+
 	r.unorderedChunks = append(
 		r.unorderedChunks[:startIdx],
 		r.unorderedChunks[startIdx+nChunks:]...)
 
 	chunkSet := newChunkSet(0, chunks[0].payloadType)
-	for _, c := range chunks {
-		chunkSet.push(c)
-	}
+	chunkSet.chunks = chunks
 
 	return chunkSet
 }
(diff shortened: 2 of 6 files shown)


reassembly_queue.go
```diff
@@ -216,15 +216,15 @@ func (r *reassemblyQueue) findCompleteUnorderedChunkSet() *chunkSet {
...
+
 	r.unorderedChunks = append(
 		r.unorderedChunks[:startIdx],
 		r.unorderedChunks[startIdx+nChunks:]...)
 
 	chunkSet := newChunkSet(0, chunks[0].payloadType)
-	for _, c := range chunks {
-		chunkSet.push(c)
-	}
+	chunkSet.chunks = chunks
```
enobufs: These lines were the root cause. `chunks` which was sliced out from `r.unorderedChunks`, was corrupted when `r.unorderedChunks` was altered. (see [this example](https://play.golang.org/p/v1qJg4DOR0K))

enobufs:
@hugo @Sean-Der took some time to figure this out. This happens when sending a large (causes SCTP level fragmentation) with unordered flag set.

pending_queue.go
```diff
@@ -65,22 +67,58 @@ func (q *pendingQueue) push(c *chunkPayloadData) {
...
 	}
 	return q.orderedQueue.get(0)
 }
 
 func (q *pendingQueue) pop(c *chunkPayloadData) error {
-	if c.unordered {
-		popped := q.unorderedQueue.pop()
-		if popped != c {
-			return fmt.Errorf("unexped chunk popped (unordered)")
+	if q.selected {
```
hugoArregui: if I have to give a summary of this selected/unorderedSelected change it will be: you need to complete your chunk before taking another one. I'm understanding it correctly?.

The problem is, IMHO, it is not so easy to reason about this code, because pop has a lot of checks regarding the current chunk but peek doesn't. So peek may give you diferent results if "selected mode" depending on wheter "selecting mode" is enabled or not, so it's assuming the caller is in a position to know that. Don't get my wrong I think this code is correct, but I wonder, if we assume the caller knows this (and relies on in), if it not may be more useful to put it all in the hands of the caller. 

For example, this is what I believe it's a similar API but all the state description (the fact that a chunk is in procress) is in the caller (association):

```
//NOTE: this is a convenience parameter because I assume is what is going to store the association
// it may also take a bool or something like that
func (q *pendingQueue) peek(cc *chunkPayloadData) *chunkPayloadData {
	if cc != nil {
		if cc.unordered {
			return q.unorderedQueue.get(0)
		}
		return q.orderedQueue.get(0)
	}

	if c := q.unorderedQueue.get(0); c != nil {
		return c
	}
	return q.orderedQueue.get(0)
}

func (q *pendingQueue) pop(c *chunkPayloadData) error {
    if c.unordered {
        popped := q.unorderedQueue.pop()
        if popped != c {
            return fmt.Errorf("unexped chunk popped (unordered)")
        }
    } else {
        popped := q.orderedQueue.pop()
        if popped != c {
            return fmt.Errorf("unexped chunk popped (ordered)")
        }
    }
	q.nBytes -= len(c.userData)
	return nil
}
```

The association will be in charge of keeping track of the current chunk and clearing it out when finished

enobufs: > if I have to give a summary of this selected/unorderedSelected change it will be: you need to complete your chunk before taking another one. I'm understanding it correctly?.

Yes.

Say if stream_A sent a message "ABC", and then stream_B sent "WXY". These are writen into pending queue after chopped into smaller chunks, {"A", "B", "C"} and then {"W", "X", "Y"}. Now, the pending queue would look like this:

pendingQueue {
  orderedQueue:  [A B C]
  unorderedQueue: [W X Y]
}

As suggested by RFC 4960 Sec 6.6, it should pick unordered chunk first if possible. Say, "A" and "W" are both beginnings of the messages, it would pick (peek or pop), "W" first. This decision must be persisted until the end of the same message, "Y".

> The problem is, IMHO, it is not so easy to reason about this code, because pop has a lot of checks regarding the current chunk but peek doesn't. So peek may give you diferent results if "selected mode" depending on wheter "selecting mode" is enabled or not, so it's assuming the caller is in a position to know that. Don't get my wrong I think this code is correct, but I wonder, if we assume the caller knows this (and relies on in), if it not may be more useful to put it all in the hands of the caller.
> 
> For example, this is what I believe it's a similar API but all the state description (the fact that a chunk is in procress) is in the caller (association):
> 
> ```
> //NOTE: this is a convenience parameter because I assume is what is going to store the association
> // it may also take a bool or something like that
> func (q *pendingQueue) peek(cc *chunkPayloadData) *chunkPayloadData {
> 	if cc != nil {
> 		if cc.unordered {
> 			return q.unorderedQueue.get(0)
> 		}
> 		return q.orderedQueue.get(0)
> 	}
> 
> 	if c := q.unorderedQueue.get(0); c != nil {
> 		return c
> 	}
> 	return q.orderedQueue.get(0)
> }
> 
> func (q *pendingQueue) pop(c *chunkPayloadData) error {
>     if c.unordered {
>         popped := q.unorderedQueue.pop()
>         if popped != c {
>             return fmt.Errorf("unexped chunk popped (unordered)")
>         }
>     } else {
>         popped := q.orderedQueue.pop()
>         if popped != c {
>             return fmt.Errorf("unexped chunk popped (ordered)")
>         }
>     }
> 	q.nBytes -= len(c.userData)
> 	return nil
> }
> ```
> 
> The association will be in charge of keeping track of the current chunk and clearing it out when finished

Ok, I need to clarify why there are peek() and pop().

When the time comes to send a new chunk, it first needs to know how much the chunk size is because all the data it would send has to be within the inflight window determined by CWND and RWND at the time. Once it learns that the chunk would fit, it will pop the chunk, more specifically, remove the one it peeked earlier from the pendingQueue.

Therefore, the implementation has the following rules:
* You can only pop chunk that peek() would return (pop() actually checks it)
* peek() does not change any state (q.selected, q.unorderedIsSelected), pop() does.
* Call sequence alway like this:

```
c := peek()
pop(c)  // c is passed to ensure what it would pop is identical to the one peeked earlier
```

Hope this helps.

hugoArregui: Great!, I think I get the idea but I don't see the problem with my suggestion, let me first complete my example (I'll simplify a TON):

```
    var currentChunk *chunkPayloadData
    for {
        c := a.pendingQueue.peek(currentChunk)
        if err := a.pendingQueue.pop(c); err != nil {
            a.log.Errorf("[%s] failed to pop from pending queue: %s", a.name, err.Error())
        }
        if !c.endingFragment {
           currentChunk = c
        } else {
           currentChunk = nil
        }

        a.inflightQueue.pushNoCheck(c)
    }
```

```
pendingQueue {
orderedQueue: [A B C]
unorderedQueue: [W X Y]
}
```

so let's say we are in initial state, let's call "ABC" chunk 1 and "WXY" chunk 2

- peek(nil) => return W
- chunk doesn't end, so current chunk 1
- peek(chunk1) => return X
- chunk doesn't end, so current chunk 1
- peek(chunk1) => return Y
- chunk doesn't end, so current chunk is nil
- peek(nil) => return A
- chunk doesn't end, so current chunk 2
- peek(chunk2) => return B
- chunk doesn't end, so current chunk 2
- peek(chunk2) => return C
- chunk doesn't end, so current chunk is nil

but all the "I'm proccesing this chunk" state is on the association. That been said this is not super important, I think your code is correct. I'll approve right now as a matter of fact

enobufs:
Thanks Hugo! I am going to work on benchmark next!

Issue #62, opened 2019-09-25
Improve data channel (SCTP) performance


enobufs:
## Motivation
SCTP (datachannel) performance is perceived very low particularly over a real network with latency with limited bandwidth. No one appears to have properly measured performance. We should identify underlying problems causing the slowness with correct measurement, then tackle those to improve it.

1. Benchmark
1. Improve performance (we may create separate tickets, and this issue will be used as an umprella ticket for those)

enobufs:
Preparation in progress. I am going to use [this tool](https://github.com/enobufs/sctptest). My focus is to measure the performance of SCTP protocol (not really about the bufferedAmount control).

enobufs:
The first pass of a benchmark on pion/sctp@v1.6.13 in comparison with TCP on various RTT (using Network Link Conditioner).

![pion_sctp@v1 6 13 and TCP](https://user-images.githubusercontent.com/930777/66262325-26943780-e792-11e9-8fd8-165d8b67e40d.png)

Obviously, the congestion window is not growing for some reason that is to be found out.

enobufs:
Further [investigation](https://docs.google.com/spreadsheets/d/1Ss69XBHSGPsbO0gXzQB1iXwNI-872e_CTPV1byyX2rY) revealed:

* TCP dynamically increases the receiver buffer size
* Pion/sctp's congestion window (cwnd) does not grow fast enough
* Retransmission receover is very very slow when massive packet loss happens

I attempted to increase the receive buffer size from 64KB (current) up to 1MB, but throughput did not improve beyond the size of 128KB.

Pion/sctp uses delayed Ack to reduce the number of Acks. Downside of it is since Ack is delayed, perceived RTT becomes large, and it would affect the overall throughput. To mitigate that, it uses "Immdidate Ack" to request the receiver to generate Ack immediately. But I found that there was a bug that prevented the immediate-ack from being sent frequently enough. By fixing this (fix-1), it improved the speed of cwnd growth a lot.

My test environment involves WiFi network, and I am seeing packet loss constantly, and sometimes a large number of packets are lost (large enought the fast-recover could not recover). When T3-Rtx timer fires, Pion/sctp reties to retransmit unack'ed packets but it was very very slow due to a bug.  By fixing it (fix-2), it was able to recover from the massive lost as fast as what TCP seems to be doing.

By having both fix-1 and fix-2, Pion/sctp was able to yield close to the theoretical throughput:

[...]

PR #79, opened 2019-10-21
Improve handling of SACK

enobufs:
Relates to #62

SCTP implementation was not following the suggestion mentioned in RFC 4960 for generating SACK, causing the poor performance in receiving data from others (Browsers, etc)

By generating SACK at every second packet, as suggested by the RFC, improved the performance to close to the theoretical throughput.

See #62 for some benchmarks.

### DIFF
--- a/ack_timer.go
+++ b/ack_timer.go
@@ -43,7 +43,7 @@ func (t *ackTimer) start() bool {
 		return false
 	}
 
-	// this is a noop if the timer is always running
+	// this is a noop if the timer is already running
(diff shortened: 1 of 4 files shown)


Sean-Der:
@enobufs this is so amazing, I can't wait to hear the results back from @chrisprobst :)

Sean-Der:
If you tag it will also create the bump PRs for you! So you don't have to wait for reviews!

enobufs:
Great. Thanks Sean!

Issue #49, opened 2019-05-28
Remove unnecessary copy during calculation of CRC32


enobufs:
## Summary
Luke pointed out on slack:
```
unrelated, but if anybody wants an ez SCTP performance optimization: https://github.com/pion/sctp/blob/master/packet.go#L147

should be creating the crc32 Hash and calling `Write` over ranges, instead of allocating/copying every single packet we send/receive (edited) 
```

## Motivation
Seems to be an easy fix/improvement.

## TODO
* Double-check if there was any intention with the copy..

AeroNotix:
I am not sure if this actually improves things:

```go
var crc32Table = crc32.MakeTable(crc32.Castagnoli)
func generatePacketChecksumNew(raw []byte) uint32 {
	chksum := crc32.New(crc32Table)
	chksum.Write(raw[:8])
	chksum.Write([]byte{0x00, 0x00, 0x00, 0x00})
	chksum.Write(raw[12:])
	return chksum.Sum32()
}

func generatePacketChecksumOld(raw []byte) uint32 {
	rawCopy := make([]byte, len(raw))
	copy(rawCopy, raw)

	// Clear existing checksum
	for offset := 8; offset <= 11; offset++ {
		rawCopy[offset] = 0x00
	}

	return crc32.Checksum(rawCopy, crc32.MakeTable(crc32.Castagnoli))
}
```

```
archlinux% go test -run=XXX -bench=. -benchtime=20s
goos: linux
goarch: amd64
pkg: github.com/pion/sctp
BenchmarkNew-12    	352351734	        71.2 ns/op
BenchmarkOld-12    	656112110	        37.6 ns/op
PASS
ok  	github.com/pion/sctp	60.301s```

kixelated:
Oh hey @AeroNotix, I didn't realize you looked into this already. Could you take a look at my pull request and see if it looks better? I can explain what was wrong with your benchmark and new implementation if you'd like.

PR #83, opened 2019-10-25
Improved generateChecksum performance. (4x)

kixelated:
The old code is egregious so I had to fix it. It would make a copy of
each packet just to unset 4 bytes. It would also create a crc32 table
each time, which is fortunately cached but requires a mutex lock.

```
benchmark                             old ns/op     new ns/op     delta
BenchmarkPacketGenerateChecksum-8     284           68.2          -75.99%
```


enobufs:
Relates to #49

AeroNotix:
@kixelated please update the commit message as per the format in the CI log and I'll be happy to merge!

AeroNotix:
@kixelated hi! Would love to merge this. Please update the commit message and I'll click that merge button.

Sean-Der:
Fantastic, thank you so much @kixelated merged as ced24b2276cf7991920a301ce5f0735cabafaf3e

Thanks for the approve @AeroNotix I just forced pushed to the branch and fixed it up :)

PR #102, opened 2020-01-31
Tear down assocation on ABORT chunk

Sean-Der:
Properly tear down the association when we receive an ABORT
chunk.

### DIFF
--- a/association.go
+++ b/association.go
@@ -426,11 +426,12 @@ func (a *Association) readLoop() {
 
 		atomic.AddUint64(&a.bytesReceived, uint64(n))
 		if err = a.handleInbound(buffer[:n]); err != nil {
-			a.log.Warnf("[%s] %s]", a.name, errors.Wrap(err, "failed to push SCTP packet").Error())
+			closeErr = err
+			break
 		}
 	}
 
-	a.log.Debugf("[%s] readLoop exited", a.name)
+	a.log.Debugf("[%s] readLoop exited %s %s", a.name, closeErr)
 }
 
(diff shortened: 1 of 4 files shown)


Sean-Der:
@enobufs I am going to write a test to confirm this doesn't regress as well, but ran out of time tonight!

association.go
```diff
@@ -1963,8 +1965,8 @@ func (a *Association) handleChunk(p *packet, c chunk) error {
 	var err error
 
 	if _, err = c.check(); err != nil {
-		return errors.Wrap(err, "failed validating chunk")
-		// TODO: Create ABORT
```
Sean-Der: I didn't find anything in the RFC, but it seems like we should tolerate anything from the other side (besides ABORT chunks)

enobufs: There are some cases mentioned in RFC 4960 which require sending ABORT  to close the association. (grep by `should abort` or `must abort` in the RFC).
Chunks that can return an error from check() currently are chunk_init.go and chunk_init_ack.go as far as I can tell.
We've been ignoring those errors so the logic hasn't be proven. I'd support you change (just log and ignore) for now to reduce regression. We could revisit this later when we finally support "sending ABORT".

Sean-Der: Ticketed! https://github.com/pion/sctp/issues/103

enobufs:
@Sean-Der Just to be sure that it can handle multiple errorCauses, I have added a marshaling test. I think your changes look good!

Sean-Der:
@enobufs thank you so much. I appreciate how much work you have put forward :)

Issue #104, opened 2020-02-01
Pion <-> usrsctp under 10% packet loss gets an abort


Sean-Der:
* Using [main.go](https://gist.github.com/d43be7f1f971f7110298f6a378efe179)  
*  [demo.js](https://gist.github.com/3fd08d4203e23e4f13b0ff6e664d94cb)  
* simulating packet loss using `comcast --device=lo --latency=150 --packet-loss=10% --target-proto=udp`

 I get an ABORT after ~5 seconds.

```
Reass 30000007,CI:ffffffff,TSN=1e5e52d3,SID=0002,FSN=1e5e52d3,SSN:0000
```

This abort is generated [here](https://github.com/sctplab/usrsctp/blob/master/usrsctplib/netinet/sctp_indata.c#L1427-L1429) in usrsctp. I am not sure why yet though, I need to read more about SCTP to fully understand.

Sean-Der:
This seems to be triggered by the size of the message we are sending. If I drop [this](https://gist.github.com/Sean-Der/d43be7f1f971f7110298f6a378efe179#file-main-go-L125) to a smaller size the crash goes away. I will keep pulling it down and see at what size it stops. (Will update this message with the data)

Sean-Der:
I don't understand what this assert is for
```
                  if ((tsn == (asoc->cumulative_tsn + 1) && (asoc->idata_supported == 0))) {
                          /* Ok we created this control and now
                           * lets validate that its legal i.e. there
                           * is a B bit set, if not and we have
                           * up to the cum-ack then its invalid.
                           */
                          if ((chk->rec.data.rcv_flags & SCTP_DATA_FIRST_FRAG) == 0) {
                                  sctp_abort_in_reasm(stcb, control, chk,
                                                      abort_flag,
                                                      SCTP_FROM_SCTP_INDATA + SCTP_LOC_7);
                                  return;
                          }
                  }
```

It seems to be saying
* TSN of this new chunk is exactly 1 greater then what we have gotten
* It doesn't have the `B` field set (beginning fragment)

Sean-Der:
Here is a `pcap` with this message `Protocol Violation: Reass 30000007,CI:ffffffff,TSN=0bf4ed02,SID=0002,FSN=bf4ed02,SSN:0000`

[pion.tar.gz](https://github.com/pion/sctp/files/4142379/pion.tar.gz)

Sean-Der:
[...]

PR #105, opened 2020-02-05
Update forward TSN to be fragment aware

Sean-Der:
When picking a new TSN value only take fragments that are at the start
or end of DATA. Taking the middle value creates a situation where we
expect the other side the re-assumable data, but we have forced them
to evict half the cache on that data already.

Relates to #104

### DIFF
--- a/association.go
+++ b/association.go
@@ -1456,16 +1456,18 @@ func (a *Association) handleSack(d *chunkSelectiveAck) error {
 		}
 
 		// RFC 3758 Sec 3.5 C2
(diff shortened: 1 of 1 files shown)


Sean-Der:
@mturnshek if you get a chance I would try this out. I for sure need a review though, this might effect things negatively I am not aware of.

enobufs:
@Sean-Der PR #107 should address this issue as discussed in #104.

Sean-Der:
Closing for the better PR #107

PR #115, opened 2020-02-25
Add synchronization to payloadQueue map access

### LINKED ISSUE #114: Race condition for chunkmap access in payloadQueue
### Your environment.
  - Version: v1.7.5
  - Browser: N/A (Snowflake)
  - Other Information - see Tor ticket [#33211](https://trac.torproject.org/projects/tor/ticket/33211)

### What did you do?
There are more details in the ticket, but we found our application going into a CPU-intensive infinite loop. Profiling pointed us towards the `markAllToRetransmit` function in payload_queue.go: https://github.com/pion/sctp/blob/master/payload_queue.go#L163

I'm not sure whether the infinite loop behaviour we saw was due to this race condition but it seems plausible.

cohosh:
Map access are not concurrency safe, and the chunkmap of a payloadQueue
is accessed by multiple goroutines. This commit adds a RWMutex to
chunkmap accesses.

#### Description

This PR is a possible solution to Issue #114. In any case, it fixes a race condition for maps that are accessed by multiple goroutines.

#### Reference issue
Fixes #114

### DIFF
--- a/payload_queue.go
+++ b/payload_queue.go
@@ -3,30 +3,36 @@ package sctp
 import (
 	"fmt"
 	"sort"
+	"sync"
 )
 
 type payloadQueue struct {
 	chunkMap map[uint32]*chunkPayloadData
+	mapMutex *sync.RWMutex
 	sorted   []uint32
 	dupTSN   []uint32
 	nBytes   int
 }
 
 func newPayloadQueue() *payloadQueue {
-	return &payloadQueue{chunkMap: map[uint32]*chunkPayloadData{}}
+	return &payloadQueue{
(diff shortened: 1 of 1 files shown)


cohosh:
Alright, I'm not sure if you still want to add this considering #110 actually fixed the issue I was having. Sounds like we weren't triggering race conditions afterall.

enobufs:
@cohosh The inflightQueue (which is a payloadQueue) is solely owned and accessed by Association, and it is protected by [Association.lock](https://github.com/pion/sctp/blob/v1.7.5/association.go#L109). I have just gone over the code to double-check that and verified that's the case. Let me close this PR.

PR #116, opened 2020-02-28
Reset retransmit flag when acked

### LINKED ISSUE #104: Pion <-> usrsctp under 10% packet loss gets an abort

enobufs:
Resolves #104

> Sorry... typo in the branch name... this is for issue 104.

Please read #104 for details.

When T3-rtx timeout occurs, it sets `retransmit` flag of all chunks in the inflightQueue to true. Immediately after that  (before the actual retransmission takes place by the writeLoop go-routne), SACK was received for some of these data chunks and their `acked` flag was set to true and userData was deleted (empty) without resetting `retransmit` flag. This ended up with sending DATA chunk with userData field being empty, causing the remote to send Abort chunk because sending DATA chunks with zero-length userData field is illegal according to RFC.

This branch reset the `retransmit` flag when SACK is received for a DATA chunk. I no longer see Abort coming back.

### DIFF
--- a/payload_queue.go
+++ b/payload_queue.go
@@ -142,6 +142,7 @@ func (q *payloadQueue) markAsAcked(tsn uint32) int {
 	var nBytesAcked int
 	if c, ok := q.chunkMap[tsn]; ok {
 		c.acked = true
+		c.retransmit = false
 		nBytesAcked = len(c.userData)
 		q.nBytes -= nBytesAcked
 		c.userData = []byte{}
--- a/payload_queue_test.go
+++ b/payload_queue_test.go
@@ -132,4 +132,28 @@ func TestPayloadQueue(t *testing.T) {
 		assert.True(t, ok, "should be true")
 		assert.True(t, c.retransmit, "should be marked as retransmit")
 	})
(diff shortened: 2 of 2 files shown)


enobufs:
Thanks @Sean-Der for prompt review!

mturnshek:
I will do a re-test today on 1.76 and report here.

PR #119, opened 2020-03-21
Stream: ReadWithContext?

gabriel:
It would be nice to have a `Stream.ReadWithContext` to allow Reads to be canceled (or timed out) via context.Context.

For example,
```go
ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
defer cancel()
buf := make([]byte, 1024)
_, err := stream.ReadWithContext(ctx, buf)
assert.EqualError(t, err, "context deadline exceeded")
```

I don't think creating a goroutine on every read is the best solution, but wanted to suggest something...

I could look into a more fully baked solution/test if this was something you might consider adding?

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -1,6 +1,7 @@
 package sctp
 
 import (
+	"context"
 	"io"
 	"math"
 	"sync"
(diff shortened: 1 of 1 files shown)


at-wat:
Yea, it's been discussed also on Go stdlib issue.
I hope that Go2 io interfaces will take `context.Context`.

Another solution fully compatible with Go1 standard interfaces is implementing `net.Conn`'s `Set(Read|Write)Deadline`. Fully implemented read/write deadline can do equivalent operations with `context.Context`, and is able to be wrapped with context interface like https://github.com/pion/dtls/tree/master/internal/net/connctx.

Sean-Der:
Hey @gabriel I am totally up for merging something like this! Sorry for my inactivity :(

gabriel:
This pull request is creating a new goroutine to wait for the context but that doesn't seem right since it might get stuck waiting for the read notify?

I think when I looked at it awhile ago I considered changing it to use channel selects everywhere instead of sync.Cond, but that seemed like too big of a change...

In my project [here](https://github.com/keys-pub/keysd/blob/master/wormhole/sctp/client.go#L79) I'm calling Read from a goroutine outside the library, so it is possible to deal with timeouts 1 call up if that makes sense.

Issue #74, opened 2019-10-01
Introduce association.Listener


enobufs:
## Summary
I found it is difficult to use pion/sctp alone because sctp.Association.Client and sctp.Association.Server take a "connected" UDP. These are equivalent to a child (server) socket and a client socket of TCP. What's missing in pion/sctp s "listening socket".
In the context of WebRTC, we use Client-Client simultaneous open, which is not practical in the cases of non-WebRTC (no-ICE, etc).  Client-Server is possible, but it would be difficult to establish connection if there's a NAT in between.

## Motivation
Introduce the usability of pion/sctp standalone in TCP style.

## Describe alternatives you've considered
I have written [a tool](https://github.com/enobufs/go-rudp) that enables TCP like listening socket capability. However, it cannot take advantage of "cookie-echo" (resembles to TCP-Syn-cookie) SCTP offers, which is a measure against DDoS attack (like SYN-flood in TCP) because internal chunk parsers are not exposed via its API.

We can use [go-rudp](https://github.com/enobufs/go-rudp) as a reference.

PR #126, opened 2020-06-20
Added DialAssociation and ListenAssociation

backkem:
These are helpers to easily run SCTP over UDP. Example adapted.
I added the Association postfix to keep space for a potential
Dial/Listen implementation that directly gives you a Stream.

Relates to #74

TODO:
* [x] Tag pion/udp after landing CI
* [ ] Change to SCTP over IP instead of over UDP.
* [ ] Proper shutdown for the listener.


backkem:
Playing around with adding Dial and Listen. However, it's not directly clear how to handle the `streamIdentifier` and `defaultPayloadType`. We could:
* Add them to the method signature or config.
* Somehow encode them into the address urls.
* Pick a default and only offer that when using these high-level methods. However, it's not clear what `defaultPayloadType` to pick.
Suggestions welcome. I may go with the 3rd option for now and just hardcode `PayloadTypeWebRTCDCEP`.

stv0g:
I would be interested in SCTP over IP as well.

@edaniels How do plan to start the work on v2? Is there a separate branch?

backkem:
@edaniels my original goal for pushing this was to make each protocol in the stack usable in their own right, not just for WebRTC. I don't have a use-case for this myself at the moment. Feel free to take this over.

edaniels:
@stv0g i think we can just cut to v2 whenever on master and if we need back ports make a v1 branch? What do you think?

edaniels:
Regarding an overall plan, don't quite have one yet but as long as we don't tag a release for it, we can be in flux API wise

edaniels:
@stv0g is making this work over IP something you'd want to take on?

JoTurk:
Closing this because I'm starting to clean SCTP, and because this is too old, maybe we can add a DetachedConn API to SCTP instead of a listener? https://github.com/pion/dtls/pull/1094 please reopen this if you're interested in rebasing / merging it.

PR #130, opened 2020-07-06
Fix handling of retransmitted cookie-echo

enobufs:
Relates to pion/webrtc#1270

#### Description
Please see this comment(https://github.com/pion/webrtc/issues/1270#issuecomment-653953743) for details.

#### What's in this PR
* Fixed the issue by sending COOKIE-ACK on the receipt of COOKIE-ECHO even in the Established state.
* Added a reproducible test using vnet, successfully confirmed the problem, and the fix.


enobufs:
@Sean-Der This is a fix for a fundamental error in handling delayed cookie-echo received in the "established" state (which would happen on lossy connection).  If you could review this at your earliest convenience, that would really be appreciated.

association.go
```diff
@@ -982,29 +982,30 @@ func (a *Association) handleHeartbeat(c *chunkHeartbeat) []*packet {
 func (a *Association) handleCookieEcho(c *chunkCookieEcho) []*packet {
 	state := a.getState()
 	a.log.Debugf("[%s] COOKIE-ECHO received in state '%s'", a.name, getAssociationStateString(state))
-	if state != closed && state != cookieWait && state != cookieEchoed {
-		return nil
-	}
+	if state != established {
+		if state != closed && state != cookieWait && state != cookieEchoed {
+			return nil
+		}
```
at-wat, line 988:
```suggestion
	switch state {
	default:
		return nil
	case established:
		break
	case closed, cookieWait, cookieEchoed:
```
may be bit readable.

at-wat: BTW, is it needed to handle cookie echo after being closed?

enobufs: Thanks @at-wat really appreciated. Good point.  Let me check the RFC...

enobufs: Yes, we need to handle Cookie Echo chunk in closed state. See [this diagram](https://github.com/pion/webrtc/issues/1270#issuecomment-650857414)

at-wat:
```
$ bash .github/lint-commit-message.sh
Update association.go

Accepted the change.

Co-authored-by: Atsushi Watanabe <atsushi.w@ieee.org>

-------------------------------------------------
The preceding commit message is invalid
it failed 'Separate subject from body with a blank line' of the following checks

* Separate subject from body with a blank line
* Limit the subject line to 50 characters
* Capitalize the subject line
* Do not end the subject line with a period
* Wrap the body at 72 characters
```
Commit message from GitHub web UI uses CR+LF which is currently not supported by our commit message lint script. https://github.com/pion/.goassets/issues/6

`git commit --amend` and remove CR will fix the error.

association.go
```diff
@@ -982,29 +982,32 @@ func (a *Association) handleHeartbeat(c *chunkHeartbeat) []*packet {
 func (a *Association) handleCookieEcho(c *chunkCookieEcho) []*packet {
 	state := a.getState()
 	a.log.Debugf("[%s] COOKIE-ECHO received in state '%s'", a.name, getAssociationStateString(state))
-	if state != closed && state != cookieWait && state != cookieEchoed {
+	switch state {
+	default:
 		return nil
-	}
+	case established:
```
at-wat, line 988: Current implementation seems returning cookie-ack whenever cookie-echo is received after established.
https://tools.ietf.org/html/rfc4960#section-5.2.4 seems to specify several conditions/actions and cookie-echo should be silently discarded in some case.

enobufs: @at-wat Yes, good finding. The current implementation does not follow the RFC for building a cookie. It is currently a random value that prevents us from validating the echoed cookie as specified in the RFC.

I am hoping to address this issue in #74 sometime soon, encoding TCB information into the cookie, then validate as the spec says. (no effect in WebRTC usage however, if we provide Listen/Dial semantics, yes it is desired.)

This issue tries to address missing state transitions from COOKIE-ECHOED to ESTABLISHED because the peer in the ESTABLISHED state does not send COOKIE-ACK, a different issue. We are seeing the connection stalled in the real world because of this.

at-wat: > I am hoping to address this issue in #74 sometime soon, encoding TCB information into the cookie, then validate as the spec says.

Great!
Maybe it's better to notify the current status of the implementation in README. If users try to use the library as a full sctp implementation, simplified implementation may cause some security problems. (as it's over DTLS in WebRTC, we don't have to care.)

enobufs: Hey @at-wat You just made me realize we at least needed to check if the echoed cookie was identical to the original one. Thanks!

Let me add these lines in the "established" case.
```
		if !bytes.Equal(a.myCookie.cookie, c.cookie) {
			return nil
		}
```

When the cookie does not match, it will silently be discarded.

enobufs: @at-wat a quick question. Use used 'break' inside the switch statement. Golang does not require break but is that something you'd recommend for readability or any other reasons? (just out of my curiosity...)

at-wat: I don't have strong reason, but I just feel it's more clear to tell that it's not `fallthrough`. (I mainly used C/C++ before starting Go)

enobufs: @at-wat I have just added the cookie comparison to the established case.

enobufs: I hear you. I write C/C++ too. ;) but I am trying to think in Go. I removed it in the last commit, hope you wouldn't mind...

at-wat: It's 100% fine to me!

PR #134, opened 2020-07-06
Fix race in rtx timer test

### LINKED ISSUE #129: Race found in travis-ci
Just from auto updating a lib https://github.com/pion/sctp/pull/128 

```
==================
WARNING: DATA RACE
Read at 0x00c000023f10 by goroutine 79:
  github.com/pion/sctp.TestRtxTimer.func4()
      /home/travis/gopath/src/github.com/pion/sctp/rtx_timer_test.go:206 +0x521
  testing.tRunner()
      /home/travis/.gimme/versions/go1.13.linux.amd64/src/testing/testing.go:909 +0x199
Previous write at 0x00c000023f10 by goroutine 78:
  github.com/pion/sctp.TestRtxTimer.func4.1()
      /home/travis/gopath/src/github.com/pion/sctp/rtx_timer_test.go:188 +0x89
  github.com/pion/sctp.(*testTimerObserver).onRetransmissionTimeout()
      /home/travis/gopath/src/github.com/pion/sctp/rtx_timer_test.go:95 +0x5a
  github.com/pion/sctp.(*rtxTimer).start.func1()
      /home/travis/gopath/src/github.com/pion/sctp/rtx_timer.go:155 +0x30d
Goroutine 79 (running) created at:
  testing.(*T).Run()
      /home/travis/.gimme/versions/go1.13.linux.amd64/src/testing/testing.go:960 +0x651
  github.com/pion/sctp.TestRtxTimer()
      /home/travis/gopath/src/github.com/pion/sctp/rtx_timer_test.go:183 +0xea
  testing.tRunner()
      /home/travis/.gimme/versions/go1.13.linux.amd64/src/testing/testing.go:909 +0x199
Goroutine 78 (running) created at:
  github.com/pion/sctp.(*rtxTimer).start()
      /home/travis/gopath/src/github.com/pion/sctp/rtx_timer.go:144 +0x1da
  github.com/pion/sctp.TestRtxTimer.func4()
      /home/travis/gopath/src/github.com/pion/sctp/rtx_timer_test.go:199 +0x38f
  testing.tRunner()
      /home/travis/.gimme/versions/go1.13.linux.amd64/src/testing/testing.go:909 +0x199
==================
```

enobufs:
#### Description
Race conditions were detected with the variable `nCbs` used in rtx_timer_tests.go.
By using sync/atomic methods on the variable, the race was gone in my local environment after updating testify to v1.6.1.

#### Reference issue
Fixes #129

### DIFF
--- a/rtx_timer_test.go
+++ b/rtx_timer_test.go
@@ -3,6 +3,7 @@ package sctp
 import (
 	//"fmt"
 	"math"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -102,10 +103,10 @@ func (o *testTimerObserver) onRetransmissionFailure(id int) {
 func TestRtxTimer(t *testing.T) {
 	t.Run("callback interval", func(t *testing.T) {
 		timerID := 0
-		var nCbs int
+		var nCbs int32
 		rt := newRTXTimer(timerID, &testTimerObserver{
 			onRTO: func(id int, nRtos uint) {
-				nCbs++
+				atomic.AddInt32(&nCbs, 1)
 				// 30 : 1 (30)
 				// 60 : 2 (90)
 				// 120: 3 (210)
@@ -126,16 +127,16 @@ func TestRtxTimer(t *testing.T) {
 		rt.stop()
 		assert.False(t, rt.isRunning(), "should not be running")
 
-		assert.Equal(t, 4, nCbs, "should be called 4 times (actual: %d)", nCbs)
+		assert.Equal(t, int32(4), atomic.LoadInt32(&nCbs), "should be called 4 times")
 	})
 
 	t.Run("last start wins", func(t *testing.T) {
 		timerID := 3
-		var nCbs int
+		var nCbs int32
 
 		rt := newRTXTimer(timerID, &testTimerObserver{
 			onRTO: func(id int, nRtos uint) {
-				nCbs++
+				atomic.AddInt32(&nCbs, 1)
 				assert.Equal(t, timerID, id, "unexpted timer ID: %d", id)
 			},
 			onRtxFailure: func(id int) {},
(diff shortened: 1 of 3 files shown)


rtx_timer_test.go
```diff
@@ -310,7 +311,7 @@ func TestRtxTimer(t *testing.T) {
 		<-doneCh
 
 		assert.True(t, rt.isRunning(), "should still be running")
-		assert.Equal(t, 6, nCbs, "must have been called 6 times")
+		assert.Equal(t, int32(6), atomic.LoadInt32(&nCbs), "should be called 6 times")
```
at-wat, line 314: As the old code caused data race with `nCbs++`, I think using atomic just replace data race error by `6 not equals to nCbs`.

enobufs: Thanks!
Btw, could you please point me to the problem? I only saw nCbs error and did not notice other problems you are takling about?

at-wat: As the race detector detected a data race, `nCbs++` and `if nCbs != expectedNCbs` in `assert.Equal` were executed concurrently. (These two memory access operations were reachable at same time from two routines.)
It should mean that `nCbs` potentially take two different values at `assert.Equal` according to the CPU load and task switch timing.

enobufs: Ah.. I see. Yes, those two events (1: nCbs++ and 2: assert.Equal) occur in two different routines but from the same cause, `onRTO` tho.  From what I can see in this particular test when onRTO callback is made, nCbs++ happens first, followed immediately by assert.Equal.  When nRtos = 4, the timeout interval would be 320msec.  If there was a delay of more than 320 msec between nCbs++ and assert.Equal, then yes, nCbs could be 7 but I think it won't happen.

enobufs: Ah...  I guess you were talking about L.206, not L.313.  Let me take a closer look at the test case.

PR #154, opened 2020-10-03
Update replace 'github.com/pkg/errors' with 'errors'

scorpionknifes:
#### Descriptions

Replace ```github.com/pkg/errors``` with ```errors```

#### Reference issue

https://github.com/pion/.goassets/issues/48


Sean-Der:
:fire: this is much better, thanks for taking on massive work again!

Can you open a PR to github.com/pion/.goassets that adds something like this to the `.golangci-lint.conf`. I haven't tried myself, but got this from https://golangci-lint.run/usage/configuration/

```
  gomodguard:
    blocked:
      modules:                                                      
        - github.com/pkg/errors:                             
```

Then after that is pushed we can tag, and will auto-open PRs across the entire org! Then we just need to go make the PRs all green like before :) most should just pass and be fine.

Thanks for taking this on again. I think it is important. Long term more people will hopefully contribute because we kept things simple.

Sean-Der:
@scorpionknifes ^ sorry forgot to tag you

Sean-Der:
Merged with 40947ace8a85f954ee88681cdacb1f20bf70c973

I also merged all the other PRs across the Pion org! You were right this is the only repo that used `pkg/errors`

If you are interested I have lots of other things that I need help with :) no pressure, but always stuff to do if you are having fun @scorpionknifes

Issue #77, opened 2019-10-07
Turn Stream.WriteSCTP() and Write()  a blocking call by default


enobufs:
### Your environment.
  - Version: v1.7.0 (or earlier)

### Background
WebRTC's datachannel.Write() does not block as we follow JavaScript WebRTC API. When sending a large amount of data, it is the application's responsibility to check the buffered amount (in SCTP layer for sending).

This is pretty standard in JavaScript land, but this really does not align with a typical semantics of Go. (i.e. net.Conn). However, implementing a flow control at the user level could be tedious and no too trivial to get it right.

### What would you expect?
We, I believe, should keep the current behavior with pion/webrtc API (maintaining JavaScript API semantics as a policy), but we could make some exceptions to it in the following cases:

* When data channel is **detached**
* When pion/datachannel or pion/sctp was used directly

In these cases, blocking Write() method can be the default behavior, and turn off the blocking when used with pion/webrtc, etc.

### What others thought
@backkem 
```
What if we add a default implementation to SCTP itself? E.g. default threshold and if you pass it stream.Write starts blocking.
If you overwrite, you're on your own.
...
Yea so, if you use SCTP or DataChannel directly -> Default blocking implementation (blocking by default seems rather idiomatic).
If you use WebRTC -> We overwrite the OnBufferedAmountLow and it disables the default implementation and otherwise confirms to the WebRTC spec. 
```

@AeroNotix 
```
that default implementation will suit 99% of people's needs
"block if exceed buffer size"
```

I totally agree with the above comments, and I think SCTP layer should take care of this.

AeroNotix:
Even after 5 years, I massively agree that the default should be to block when there is too much buffered data in-flight.

enobufs:
I still have the same feeling, even more strongly also!
When we do blocking write, I feel we should ditch the 'buffered amount' flow control mechanism (it's for javascript and it's been the source of confusion for Go programmers!) from supporting both modes. (thoughts?)
Let us plan for pion/sctp v2.

Sean-Der:
Should we make it part of pion/webrtc@v4 ? We can start a v2 now!

edaniels:
Do we think a blocking write would be a write that has made it to inflight and that there's no longer a pending queue? That could simplify/remove some code.

[...]

PR #176, opened 2020-12-20
Implement shutdown sequence

### LINKED ISSUE #152: Implement Shutdown Sequence
Currently `Close()` doesn't implement the shutdown sequence.

jeremija:
#### Description
I finally had some time to continue the work on adding the multi server node support on `Peer Calls`. I noticed there was an issue where the other side did not notice that a SCTP association was closed from the other side.

I tried to follow the steps from the [`SCTP Association State Diagram`][1] and [`Shutdown of an Association`][2] sections of [RFC 4960][3].

I added two new tests and all existing tests passed.

Instead of modifying the existing `Close` function I added a new `Shutdown` method to keep the backwards compatibility. I also thought it made sense to allow the users to pick which teardown strategy to use.

Perhaps the `Close` should be modified to use `Abort`?

#### Reference issue
Fixes #152.

[1]: https://tools.ietf.org/html/rfc4960#section-4
[2]: https://tools.ietf.org/html/rfc4960#section-9.2
[3]: https://tools.ietf.org/html/rfc4960

### DIFF
--- a/association.go
+++ b/association.go
@@ -2,6 +2,7 @@ package sctp
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -19,6 +20,7 @@ import (
 var (
 	globalMathRandomGenerator        = randutil.NewMathRandomGenerator() // nolint:gochecknoglobals
 	errChunk                         = errors.New("abort chunk, with following errors")
+	errShutdownNonEstablished        = errors.New("shutdown called in non-established state")
 	errAssociationClosedBeforeConn   = errors.New("association closed before connecting")
 	errSilentlyDiscard               = errors.New("silently discard")
 	errInitNotStoredToSend           = errors.New("the init not stored to send")
@@ -65,6 +67,7 @@ const (
 const (
 	timerT1Init int = iota
 	timerT1Cookie
+	timerT2Shutdown
 	timerT3RTX
 	timerReconfig
 )
@@ -146,6 +149,10 @@ type Association struct {
 	willRetransmitFast     bool
 	willRetransmitReconfig bool
 
+	willSendShutdown         bool
+	willSendShutdownAck      bool
+	willSendShutdownComplete bool
+
 	// Reconfig
 	myNextRSN        uint32
 	reconfigs        map[uint32]*chunkReconfig
@@ -178,12 +185,13 @@ type Association struct {
 	fastRecoverExitPoint uint32
 
 	// RTX & Ack timer
-	rtoMgr    *rtoManager
-	t1Init    *rtxTimer
-	t1Cookie  *rtxTimer
-	t3RTX     *rtxTimer
-	tReconfig *rtxTimer
-	ackTimer  *ackTimer
+	rtoMgr     *rtoManager
+	t1Init     *rtxTimer
+	t1Cookie   *rtxTimer
+	t2Shutdown *rtxTimer
+	t3RTX      *rtxTimer
+	tReconfig  *rtxTimer
+	ackTimer   *ackTimer
 
 	// Chunks stored for retransmission
 	storedInit       *chunkInit
@@ -317,8 +325,9 @@ func createAssociation(config Config) *Association {
 
 	a.t1Init = newRTXTimer(timerT1Init, a, maxInitRetrans)
 	a.t1Cookie = newRTXTimer(timerT1Cookie, a, maxInitRetrans)
-	a.t3RTX = newRTXTimer(timerT3RTX, a, noMaxRetrans)        // retransmit forever
-	a.tReconfig = newRTXTimer(timerReconfig, a, noMaxRetrans) // retransmit forever
+	a.t2Shutdown = newRTXTimer(timerT2Shutdown, a, noMaxRetrans) // retransmit forever
+	a.t3RTX = newRTXTimer(timerT3RTX, a, noMaxRetrans)           // retransmit forever
+	a.tReconfig = newRTXTimer(timerReconfig, a, noMaxRetrans)    // retransmit forever
 	a.ackTimer = newAckTimer(a)
 
 	return a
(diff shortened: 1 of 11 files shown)


Sean-Der:
Oh wow this is amazing! Nice work @jeremija sorry I got distracted by `v3.0.0` :(

Reviewing now @enobufs would love your help (if you can!) you know this way better then me.

enobufs:
My apologies for not being able spare time for this. I will review this from now.  I have just rebased on current master. It appears to be such an excellent work indeed! :D

association.go
```diff
@@ -393,18 +401,40 @@ func (a *Association) sendCookieEcho() error {
...
+	if a.inflightQueue.size() == 0 {
+		// No more outstanding, send shutdown.
+		a.willSendShutdown = true
+		a.awakeWriteLoop()
+		a.setState(shutdownSent)
+	}
+
+	a.lock.Unlock()
+
+	return a.closeWriteLoopCh, nil
```
enobufs, line 430: I am a bit nervous about returning the channel.  May I hear your thoughts on this?

enobufs (commented):
This is really an amazing work!  Thanks so much for being one of the very few contributors of SCTP.  I have left some comments. Also, let me share [my notes](https://docs.google.com/document/d/16ciW1swfX0I8jfLqMrj4v6_FxNUQbgX5IwMbZAuYH6Q) just if case it helps.
Thanks again!

jeremija:
Thanks for your review. No worries about not being about the time - this is open source after all :) I've lost a little bit of context, but I'll do my best to answer your questions  :slightly_smiling_face:

association.go
jeremija: I think I just wanted a way for the caller to know when the shutdown process has completed since it is does not happen in an instant like `Close` does.

association.go
```diff
@@ -1537,15 +1662,81 @@ func (a *Association) handleSack(d *chunkSelectiveAck) error {
...
+	state := a.getState()
+
+	switch state {
+	case established:
+		if a.inflightQueue.size() > 0 {
+			a.setState(shutdownReceived)
+		} else {
+			// No more outstanding, send shutdown ack.
+			a.willSendShutdownAck = true
+			a.t2Shutdown.start(a.rtoMgr.getRTO())
```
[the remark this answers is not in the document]
jeremija: Oh I think you're right, I removed it and the tests still passed.

association.go
```diff
@@ -1537,15 +1662,81 @@ func (a *Association) handleSack(d *chunkSelectiveAck) error {
...
+			a.setState(shutdownAckSent)
+
+			a.awakeWriteLoop()
+		}
+
+		// a.cumulativeTSNAckPoint = c.cumulativeTSNAck
+	case shutdownSent:
+		a.willSendShutdownAck = true
+		a.t2Shutdown.start(a.rtoMgr.getRTO())
```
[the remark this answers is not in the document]
jeremija: Right - I removed it and the tests still passed.

association.go
```diff
@@ -1537,15 +1662,81 @@ func (a *Association) handleSack(d *chunkSelectiveAck) error {
...
+		a.setState(shutdownSent)
+	case state == shutdownReceived:
+		// No more outstanding, send shutdown ack.
+		shouldAwakeWriteLoop = true
+		a.willSendShutdownAck = true
+		a.setState(shutdownAckSent)
+	}
+
+	if shouldAwakeWriteLoop {
+		a.awakeWriteLoop()
```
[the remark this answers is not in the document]
jeremija: Hmm, I don't think I changed this behavior. The `shouldAwakeWriteLoop` argument in `sctp.Association.postprocessSack` is taken from the value of `cumTSNAckPointAdvanced` at [association.go:1665](https://github.com/pion/sctp/blob/85af7427b4a18641424f1b119bfd675b3e500d36/association.go#L1665) at the end of the `sctp.Association.handleSack` method.

So previously we had:

```go	
func (a *Association) handleSack(d *chunkSelectiveAck) error {
	// ... 

	if a.inflightQueue.size() > 0 {
		// Start timer. (noop if already started)
		a.log.Tracef("[%s] T3-rtx timer start (pt3)", a.name)
		a.t3RTX.start(a.rtoMgr.getRTO())
	}

	if cumTSNAckPointAdvanced {
		a.awakeWriteLoop()
	}

	return nil
}
```

And now we have:

```go
func (a *Association) handleSack(d *chunkSelectiveAck) error {
	// ... 

	a.postprocessSack(state, cumTSNAckPointAdvanced)

	return nil
}

// The caller must hold the lock. This method was only added because the
// linter was complaining about the "cognitive complexity" of handleSack.
func (a *Association) postprocessSack(state uint32, shouldAwakeWriteLoop bool) {
	switch {
	case a.inflightQueue.size() > 0:
		// Start timer. (noop if already started)
		a.log.Tracef("[%s] T3-rtx timer start (pt3)", a.name)
		a.t3RTX.start(a.rtoMgr.getRTO())
	case state == shutdownPending:
		// No more outstanding, send shutdown.
		shouldAwakeWriteLoop = true
		a.willSendShutdown = true
		a.setState(shutdownSent)
	case state == shutdownReceived:
		// No more outstanding, send shutdown ack.
		shouldAwakeWriteLoop = true
		a.willSendShutdownAck = true
		a.setState(shutdownAckSent)
	}

	if shouldAwakeWriteLoop {
		a.awakeWriteLoop()
	}
}
```

If you still have any concerns, it's possible that I missed something here.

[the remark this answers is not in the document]
enobufs: Oh I see, my bad.  It looks good to me now.

association.go
enobufs: Shutdown sequence (esp. graceful one)  is always a headache... 
I personally feel that the Shutdown() should block on the channel by itself.  If necessary, caller can always choose to call Shutdown() from a goroutine for async operation. Hmm, let me ask Sean.

@Sean-Der What is thought on this? The Shutdown() method triggers the shutdown sequence. Should it block until the sequence is complete, or return a channel so that the caller could choose to wait on the channel...?
Pion mixes Javascript (non-blocking) and Go (blocking) styles which confuses myself sometimes...

Sean-Der: I think blocking is the way to go. You can optionally pass in a context to be notified. I am not an expert Go developer either though :)

* [crypto/tls](https://golang.org/pkg/crypto/tls/#Conn.Close) is a blocking Close
* HTTP provides two APIs [Context](https://golang.org/pkg/net/http/#Server.Shutdown) and [Blocking](https://golang.org/pkg/net/http/#Server.Close)
* [net.Conn](https://golang.org/pkg/net/#Conn) is blocking as well.

---

So excited to have you both back I missed working together! I learn so much from your PRs and so fun to see Pion progress so quickly without me :)

jeremija: Thanks, I like the `context.Context` way. I just pushed an update that modifies the `sctp.Association.Shutdown`'s signature to:

```go
// Shutdown initiates the shutdown sequence. The method blocks until the
// shutdown sequence is completed and the connection is closed, or until the
// passed context is done, in which case the context's error is returned.
func (a *Association) Shutdown(ctx context.Context) error {
```

Does that work?

jeremija: > So excited to have you both back I missed working together! I learn so much from your PRs and so fun to see Pion progress so quickly without me :)

Aww thanks :heart: I miss Pion, I learned so much from the whole community and I don't think I'd have the clients I currently have if it weren't for you guys!

enobufs (approved):
@Sean-Der thanks for your comment. @jeremija I like the use of context too, more idiomatic in Go. I can also see that when we improve the detached datachannel (#77), and introduction of Listen/Accept (#74), this Shutdown method with context will become very useful. At that point, as @jeremija pointed out, we'd consider renaming Shutdown to Close and the current Close to Abort, etc. Let me approve this.

jeremija:
Thanks for your review @enobufs!

> we'd consider renaming Shutdown to Close and the current Close to Abort

Actually I think the current implementation of `Close` is not implemented as per the specification for `Abort`:

```
9.1.  Abort of an Association

   When an endpoint decides to abort an existing association, it MUST
   send an ABORT chunk to its peer endpoint.  The sender MUST fill in
   the peer's Verification Tag in the outbound packet and MUST NOT
   bundle any DATA chunk with the ABORT.  If the association is aborted
   on request of the upper layer, a User-Initiated Abort error cause
   (see Section 3.3.10.12) SHOULD be present in the ABORT chunk.

   An endpoint MUST NOT respond to any received packet that contains an
   ABORT chunk (also see Section 8.4).

   An endpoint receiving an ABORT MUST apply the special Verification
   Tag check rules described in Section 8.5.1.

   After checking the Verification Tag, the receiving endpoint MUST
   remove the association from its record and SHOULD report the
   termination to its upper layer.  If a User-Initiated Abort error
   cause is present in the ABORT chunk, the Upper Layer Abort Reason
   SHOULD be made available to the upper layer.
```

According to my understanding, we don't send any chunk after `Close` is called, we just exit the write goroutine. But I might as well be missing something here.

PR #183, opened 2021-02-20
Add sctp.Association.Abort(reason string) method

### LINKED ISSUE #182: Implement Abort according to RFC 4960
Currently the `Close` methods immediately closes the connection.

We should send an abort packet before closing the connection.

jeremija:
According to [RFC 4960 Section 9.1](https://tools.ietf.org/html/rfc4960#section-9.1). Also see my comment at #176, more precisely: https://github.com/pion/sctp/pull/176#issuecomment-782578388

Closes #182.

I don't think we currently do any tag verification for any packets, but we can implement that later.

I _think_ that `Close` method should be modified to call `Abort("close")` or similar; I was unable to find any part of the spec that allows just closing the underlying connection.

### DIFF
--- a/error_cause_user_initiated_abort.go
+++ b/error_cause_user_initiated_abort.go
@@ -1,0 +1,46 @@
+package sctp
+
+import (
+	"fmt"
+)
+
+/*
+   This error cause MAY be included in ABORT chunks that are sent
+   because of an upper-layer request.  The upper layer can specify an
+   Upper Layer Abort Reason that is transported by SCTP transparently
+   and MAY be delivered to the upper-layer protocol at the peer.
+
+        0                   1                   2                   3
+        0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+       |         Cause Code=12         |      Cause Length=Variable    |
+       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+       /                    Upper Layer Abort Reason                   /
+       \                                                               \
+       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+*/
+type errorCauseUserInitiatedAbort struct {
+	errorCauseHeader
+	upperLayerAbortReason []byte
+}
+
+func (e *errorCauseUserInitiatedAbort) marshal() ([]byte, error) {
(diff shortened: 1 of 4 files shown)


error_cause_user_initiated_abort.go
```diff
@@ -0,0 +1,32 @@
+package sctp
+
+import (
+	"fmt"
+)
+
```
enobufs, line 6: Are you able to add the following text just to be consistent with other error_cause_xxx.go files?
```
/*
   This error cause MAY be included in ABORT chunks that are sent
   because of an upper-layer request.  The upper layer can specify an
   Upper Layer Abort Reason that is transported by SCTP transparently
   and MAY be delivered to the upper-layer protocol at the peer.

        0                   1                   2                   3
        0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
       |         Cause Code=12         |      Cause Length=Variable    |
       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
       /                    Upper Layer Abort Reason                   /
       \                                                               \
       +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
*/
```

enobufs (changes requested):
Just one minor request from me. There's merge conflict to resolve. Other changes look good to me.

error_cause_user_initiated_abort.go
jeremija: Sure, I'll try to address it over the next couple of days

jeremija: Done!

enobufs: Thanks @jeremija !

Some auto tests are failing with TestAssociation_HandlePacketBeforeInit/Abort.
From what I can see, a.handleAbort() calls a.close() which sets the state to `closed`. Then the test case calls a.Close() which complains "use of closed network connection".

I think calling a.close() on the receipt of Abort chunk. I believe we could skip a.Close() when the test case is `"Abort"` in TestAssociation_HandlePacketBeforeInit.

Otherwise, it all looks good to me!

enobufs: This is something from before but I think this test case name should be changed to something like: "TestAssociation_HandlePacketInCookieWaitState" IMO.

enobufs:
@jeremija I am pretty sure this is a test issue. If you'd prefer, I can sort this out for you. Please let me know.

error_cause_user_initiated_abort.go
jeremija: Thanks, I didn't realize there was a failing test. It should be fixed now. I also renamed the test case.

enobufs (approved):
Thanks @jeremija . LGTM!

PR #205, opened 2021-10-06
Add more functionality for SCTPTransportStats

wawesomeNOGUI:
#### Description
Adds functions to get the stats required in  https://www.w3.org/TR/webrtc-stats/#sctptransportstats-dict*

Also is there a way to get the amount of unacked data right now? I couldn't find a function/ parameter for it yet


stv0g:
I am closing this in favour of #263

PR #207, opened 2021-11-08
Add Assocation.GetOrOpenStream

edaniels:
This allows negotiated channels to be fetched when created
out of band of SDP.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -8,6 +8,7 @@ Atsushi Watanabe <atsushi.w@ieee.org>
 backkem <mail@backkem.me>
 Cecylia Bocovich <cohosh@torproject.org>
 chenkaiC4 <chenkaic4@gmail.com>
+Eric Daniels <eric@erdaniels.com>
 Hugo Arregui <hugo.arregui@gmail.com>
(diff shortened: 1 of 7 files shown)


Sean-Der:
@edaniels Is this because we have a race of both side creating the stream with the same id?

LGTM! I tempted to remove the check from `OpenStream` and have it return the stream if it is already opened to keep the API simple.

edaniels:
I'm cool with removing the check if no one else is using the package in a manner where it should be a failure if it's already open (although we would need to bump the major version, right?). If someone was using SCTP outside of the context of WebRTC, they may want to error out opening twice if say their library, if it didn't do this, caused it start double reading/writing.

Sean-Der:
@edaniels Good point. I guess better safe than sorry.

My last concern is taking the lock for getting the ppid. Do you think it will add significant lock contention? Would it be worth making it an atomic?

Sean-Der (approved):

Sean-Der:
Your call. What do you think will give us the most maintainable code long term?

edaniels:
Yeah let's make the simplification and add the atomic.

PR #241, opened 2022-10-15
stream: Add "SetReadDeadline"

### LINKED ISSUE #240: SCTP Streams have no SetReadDeadline
## Summary

SCTP Streams have no SetReadDeadline and so timeouts cannot be implemented

## Motivation

Most connections (net.Conn, net.PacketConn) in Go allow "SetReadDeadline" so a read timeout can be implemented by the caller.

## Describe alternatives you've considered

Application can reimplement using a separate receive goroutine and channels, but it would be nice to support this in the library

nerd2:
#### Description

Add a read-deadline method similar to net.Conn/net.PacketConn to allow read timeouts to be implemented

#### Reference issue
Fixes #240

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"os"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/pion/logging"
 )
@@ -46,6 +48,7 @@ func (ss StreamState) String() string {
 var (
 	errOutboundPacketTooLarge = errors.New("outbound packet larger than maximum message size")
 	errStreamClosed           = errors.New("stream closed")
+	errReadDeadlineExceeded   = fmt.Errorf("read deadline exceeded: %w", os.ErrDeadlineExceeded)
 )
 
 // Stream represents an SCTP stream
@@ -58,6 +61,7 @@ type Stream struct {
 	sequenceNumber      uint16
 	readNotifier        *sync.Cond
 	readErr             error
+	readTimeoutCancel   chan struct{}
 	unordered           bool
 	reliabilityType     byte
 	reliabilityValue    uint32
@@ -115,6 +119,14 @@ func (s *Stream) ReadSCTP(p []byte) (int, PayloadProtocolIdentifier, error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
+	defer func() {
+		// close readTimeoutCancel if the current read timeout routine is no longer effective
+		if s.readTimeoutCancel != nil && s.readErr != nil {
+			close(s.readTimeoutCancel)
+			s.readTimeoutCancel = nil
+		}
+	}()
+
 	for {
 		n, ppi, err := s.reassemblyQueue.read(p)
 		if err == nil {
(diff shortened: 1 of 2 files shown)


stream.go
```diff
@@ -132,6 +136,43 @@ func (s *Stream) ReadSCTP(p []byte) (int, PayloadProtocolIdentifier, error) {
...
+		if errors.Is(s.readErr, errReadDeadlineExceeded) {
+			s.readErr = nil
+		}
+	} else {
+		s.readTimeoutCancel = make(chan struct{})
+
+		go func(readTimeoutCancel chan struct{}) {
+			t := time.NewTimer(time.Until(deadline))
+			select {
+			case <-readTimeoutCancel:
```
enobufs, line 159: I think this channel should be closed when s.readErr is set to a non-nil value outside this function so that this goroutine would exit right away. Any thoughts?

enobufs (commented):
I have left one comment. Everything else looks good, vnice work!

stream.go
nerd2: Agreed, updated. Thanks!

enobufs: Thanks @nerd2!

I have just noticed there is [one other path](https://github.com/nerd2/sctp/blob/396e5d548cb3fdecec4f2dc02e19b60c5c842286/association.go#L606) that sets s.readError to a non-nil value.

I'm wondering if we could close the channel in one place using defer inside s.ReadSCTP method...

```go
func (s *Stream) ReadSCTP(p []byte) (int, PayloadProtocolIdentifier, error) {
	s.lock.Lock()
	defer s.lock.Unlock()

        defer func() {
                // close readTimeoutCancel if not nil while the lock is still held
                if s.readTimeoutCancel != nil {
                        close(s.readTimeoutCancel)
                        s.readTimeoutCancel = nil
                }
        }()

	for {
```

nerd2: I believe even if a single (or multiple) ReadSCTP succeeds, the definition of SetReadDeadline requires the read timeout routine still to run so it applies to following ReadSCTPs.

BTW I did update association.go too

enobufs: Yes, I agree. As I forgot to add `&& s.readErr != nil` to the if condition in my code snippet, let me ask your thought once again.

```go
func (s *Stream) ReadSCTP(p []byte) (int, PayloadProtocolIdentifier, error) {
	s.lock.Lock()
	defer s.lock.Unlock()

        defer func() {
                // close readTimeoutCancel if the current read timeout routine is no longer effective
                if s.readTimeoutCancel != nil && s.readErr != nil {
                        close(s.readTimeoutCancel)
                        s.readTimeoutCancel = nil
                }
        }()

	for {
```

Note:
* If s.readErr != nil, subsequent ReadSCTPs would return immediately with the error

stream.go
```diff
@@ -132,6 +136,43 @@ func (s *Stream) ReadSCTP(p []byte) (int, PayloadProtocolIdentifier, error) {
...
+func (s *Stream) SetReadDeadline(deadline time.Time) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if s.readTimeoutCancel != nil {
+		close(s.readTimeoutCancel)
+		s.readTimeoutCancel = nil
+	}
+
+	if deadline.IsZero() {
```
enobufs, line 149: I feel if s.readErr != nil && s.readErr != errReadDeadlineExceeded, we could just return (probably with nil).

enobufs: Example:

```go
// SetReadDeadline sets the read deadline in an identical way to net.Conn
func (s *Stream) SetReadDeadline(deadline time.Time) error {
	s.lock.Lock()
	defer s.lock.Unlock()

	if s.readTimeoutCancel != nil {
		close(s.readTimeoutCancel)
		s.readTimeoutCancel = nil
	}

	if s.readErr != nil {
		if !errors.Is(s.readErr, errReadDeadlineExceeded) {
			return nil	
		}
		s.readErr = nil
	}

	s.readTimeoutCancel = make(chan struct{})

	go func(readTimeoutCancel chan struct{}) {
		// as-is
	}
	return nil
}
```

stream.go
nerd2: Added, and removed readTimeoutCancel handling from other readErr cases

stream.go
nerd2: Agreed, added

enobufs (approved):
Thanks so much @nerd2. SetReadDeadline() is going to be very useful.

Issue #239, opened 2022-10-13
Remove the pending buffer after the data channel call close() by either side.


jerry-tao:
## Summary

Remove the pending buffer after the channel is closed by either side.

## Motivation

Please refer to #238 

## Describe alternatives you've considered

In the official WebRTC code, all pending data will be clear after the data channel call close() by either side.

But the SCTP RFC does not suggest so.  The chrome will ACK the closing channel data too.

Based on @enobufs 's work, here is my thought:

- If we call close or receive reconfig, we should transform to closing state.
- When the close reconfig is finished, we transform to closed state.
- During the closing/closed state, the read and write should be rejected with error.
- During the closing/closed state, stop sending buffered data in pendingQueue.

Or we could emit the closing state to the caller ( data channel) to decide what to do.

```
dc.OnClosing(func(){
   dc.Close()
})
```

## Additional context

During reading the official webrtc code, I found it has more assumptions like: 

- Default buffer size for one data channel is 16MB, more data will be rejected.
- The max data channel count is 1024, rather not 65535.

## References

- #238 
- [Issue 559394: Closing WebRTC data channels is broken and only works chrome to chrome
](crbug.com/559394)

enobufs:
> In the official WebRTC code, all pending data will be clear after the data channel call close() by either side.

Not true.
W3C spec says in [6.2.4 Closing procedure](https://www.w3.org/TR/webrtc/#closing-procedure):
```
1. Finish sending all currently pending messages of the channel.
2. Follow the closing procedure defined for the channel's underlying data transport :
    1. In the case of an SCTP-based transport, follow [RFC8831], section 6.7.
```

But `Chrome` does not seem to bother sending the pending data **when the remote data channel has already called "close()"**.

jerry-tao:
@enobufs You are right, I made a mistake in my test code.

So the only change should be:

- If received reconfigs with outgoingReset, clear sending pending buffer.

enobufs:
I found myself still debating whether the pending buffer should be discarded OutgoingResetRequest was already received. Remote sent OutgoingResetRequest ("I am not going to send any more") but not IncomingResetRequest ("please stop sending any more"). I can see RFC8831/6252 had a graceful shutdown (e.g. shutdown(SHUT_WR)) in mind.

> shutdown(SHUT_WR) == OutgoingResetRequest

Interoperability with browsers is important but Pion does not need to follow everything what W3C spec says. There may be something Pion could take advantage of being a native library...

[...]

PR #245, opened 2022-11-04
Remove pending buffer when stream closed

jerry-tao:
#### Description

See #239 

#### Reference issue

See #239

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -24,12 +24,12 @@ const (
 
 // StreamState is an enum for SCTP Stream state field
 // This field identifies the state of stream.
-type StreamState int
+type StreamState int32
 
 // StreamState enums
 const (
 	StreamStateOpen    StreamState = iota // Stream object starts with StreamStateOpen
-	StreamStateClosing                    // Outgoing stream is being reset
+	StreamStateClosing                    // Stream is closed by remote
 	StreamStateClosed                     // Stream has been closed
 )
 
@@ -71,6 +71,7 @@ type Stream struct {
 	state               StreamState
 	log                 logging.LeveledLogger
 	name                string
+	version             uint32
 }
 
 // StreamIdentifier returns the Stream identifier associated to the stream.
@@ -296,6 +297,7 @@ func (s *Stream) packetize(raw []byte, ppi PayloadProtocolIdentifier) []*chunkPa
 		copy(userData, raw[i:i+fragmentSize])
 
 		chunk := &chunkPayloadData{
+			streamVersion:        s.version,
 			streamIdentifier:     s.streamIdentifier,
 			userData:             userData,
 			unordered:            unordered,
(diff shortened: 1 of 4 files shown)


stream.go
```diff
@@ -28,9 +28,10 @@ type StreamState int
 
 // StreamState enums
 const (
-	StreamStateOpen    StreamState = iota // Stream object starts with StreamStateOpen
-	StreamStateClosing                    // Outgoing stream is being reset
-	StreamStateClosed                     // Stream has been closed
+	StreamStateOpen          StreamState = iota // Stream object starts with StreamStateOpen
+	StreamStateClosing                          // Outgoing stream is being reset
```
enobufs, line 32: As [discussed earlier](https://github.com/pion/sctp/issues/239#issuecomment-1284985062), stream.Close() should close both inbound and outbound. (the state will immediately goes to StreamStateClosed)

When s.Close() is called, we should set [s.readErr to io.EOF](https://github.com/pion/sctp/blob/d0b7cf3963e99337bdf09e11989cf8b0916e1179/stream.go#L282) then [call s.readNotifier.Broadcast()](https://github.com/pion/sctp/blob/d0b7cf3963e99337bdf09e11989cf8b0916e1179/stream.go#L286) as we used to do.

enobufs (commented):
Removal of pending data looks good. I have one comment. Please let me know if you have any questions.

stream.go
jerry-tao: @enobufs  Updated.

enobufs: @jerry-tao Are you interested in fixing the failing tests?  If not, I can help you this weekend.

jerry-tao: I'm working on the failing test, I will commit the code later.
But in our browser test, it failed again and stuck at closing, working on it now.

vnet_test.go
```diff
@@ -529,16 +535,17 @@ func TestStreamClose(t *testing.T) {
 			venv.dropNextReconfigChunk(1)
 		}
 
-		// Immediately close the stream
+		// Wait server accept stream.
+		time.Sleep(time.Millisecond * 100)
```
jerry-tao, line 539: @enobufs Here I have another problem, if close the client stream Immediately, the server won't receive anything, it will block at server.AcceptStream(), block the test.
I think it's fine in the real world, do you think we should let DECP pass before closed?

enobufs: Please review this PR.  I simply reverted the changes to TestStreamClose.
https://github.com/jerry-tao/sctp/pull/4

jerry-tao: @enobufs  Thanks for your help, I was planning to fix CI when the PR is ready. 
I'm still testing it with browser, need one more day to make sure it worked as expected.
When the PR is ready, I will remove the WIP title.

jerry-tao:
@enobufs I think this PR is ready to merge, but there are two minor problems left:
- If close too quickly, the DECP message may be removed too.
- When the buffer is big, the sisToReset message will be delayed.

I think these could be fixed when refactoring as you said in #77.

enobufs (approved):
I'm glad that we came to a solution we both agree. Thanks @jerry-tao for your support.

PR #249, opened 2022-11-20
Use new pion/transport Net interface

stv0g:
This change adapts pion/ice to use a new interface for most network related operations. The interface was formerly a simple struct vnet.Net which was originally intended to facilicate testing. By replacing it with an interface we have greater flexibility and allow users to hook into the networking stack by providing their own implementation of the interface.

See https://github.com/pion/transport/pull/204 for a detailed description

See also #248


enobufs (approved):
Thans @stv0g. lgtm!

PR #254, opened 2022-12-12
Send enqueued chunks when in closing state

edaniels:
This slightly scales back the work done in https://github.com/pion/sctp/pull/245 which appears to have caused a regression in what https://github.com/pion/webrtc appears to expect in terms of data channel closing behavior with respect to pending writes on the sender side. More specifically, `TestEOF/Detach` started to fail where we expected that when the sender did a write followed by close, that the reader would see those writes and then see an EOF.

@enobufs, to be honest, I'm not 100% sure if the original PR was correct or if I'm even correct here. The change I made here is to still follow [RFC6525](https://www.rfc-editor.org/rfc/rfc6525#section-5.1) for outgoing resets where we wait to send queued reads.

### DIFF
--- a/association.go
+++ b/association.go
@@ -124,12 +124,13 @@ func getAssociationStateString(a uint32) string {
 //
 // Tag         :
 // State       : A state variable indicating what state the association
-//             : is in, i.e., COOKIE-WAIT, COOKIE-ECHOED, ESTABLISHED,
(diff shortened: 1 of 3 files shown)


edaniels:
This actually doesn't fully cover it. Going to open an issue.

PR #255, opened 2022-12-12
Revert "Remove pending buffer when stream closed"

edaniels:
This reverts commit c0159aa2d49c240362038edf88baa8a9e6cfcede.

@jerry-tao @enobufs, the reverted commit made sense in theory but it's causing a regression in what https://github.com/pion/webrtc appears to expect in terms of data channel closing behavior with respect to pending writes on the sender side. More specifically, `TestEOF/Detach` started to fail where we expected that when the sender did a write followed by close, that the reader would see those writes and then see an EOF. That is to say, the enqueued writes are lost due to these changes. It doesn't seem like this change fixed any correctness issues with regards to the spec so it seems like we should revert it so as to unblock other changes in this v1.8.x. v1.8.4 is currently published but is now broken/flaky. https://github.com/pion/sctp/pull/254 attempted to fix this but it introduced more issues due to how we lose a lot of context when we `Close` in that we don't know why we are closing (inbound reset or user requested).

### DIFF
--- a/association.go
+++ b/association.go
@@ -221,9 +221,8 @@ type Association struct {
 	delayedAckTriggered   bool
 	immediateAckTriggered bool
 
-	name          string
-	log           logging.LeveledLogger
-	streamVersion uint32
+	name string
+	log  logging.LeveledLogger
 }
 
 // Config collects the arguments to createAssociation construction into
@@ -1369,7 +1368,6 @@ func (a *Association) createStream(streamIdentifier uint16, accept bool) *Stream
 		streamIdentifier: streamIdentifier,
 		reassemblyQueue:  newReassemblyQueue(streamIdentifier),
 		log:              a.log,
-		version:          atomic.AddUint32(&a.streamVersion, 1),
 		name:             fmt.Sprintf("%d:%s", streamIdentifier, a.name),
 	}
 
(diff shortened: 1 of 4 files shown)


edaniels:
I reverted for one commit and back in `master` to unblock other feature work.

edaniels:
If we're okay with how this is breaking the `webrtc` tests, then I think we should just go update the `TestEOF/Detach` test in `datachannel_go_test.go` to use channel synchronization to wait for the read of `this is some test data` to happen prior to the sender side closing; otherwise the test will be flaky.

enobufs (approved):
Hi @edaniels I wasn't aware v1.8.4 caused an error at pion/webrtc. I should have tested it earlier. I agree, we should revert it for now. I can see you have already done it with v1.8.5 - nice. I will revisit this and look into the affected test later.

jerry-tao:
LGTM.

jerry-tao (approved):

jerry-tao:
As we discussed in #245, the commit does not fully resolve the origin issue either. As I said in #239 , the pion/sctp is more like webrtc/sctp_transport+webrtc/datachannel+libsrtcp.
Maybe in v2.0 we could split the sctp_transport and datachannel logic to pion/datachannel.
As for the correctness, this bothers me too, should we have the same logic with C++ webrtc/src or implements it according to RFC with personal understanding?

edaniels:
Yeah I think splitting it up to more closely reflect the C++ code modulo idioms would be good. I think the really important bit is what @enobufs inquired about with the RFC coauthor regarding close in the W3C API

PR #286, opened 2023-08-08
Use IANA assigned UDP port number for SCTP/UDP

tuexen:
#### Description
This allows Wireshark to detect SCTP automatically.


Sean-Der:
I feel super honored to have you look at pion/sctp @tuexen!

hopefully we didn’t get anything too wrong :)

PR #298, opened 2024-02-08
Support Unknown Parameters in INIT Chunk

Sean-Der:
Start to implement section-3.2.1 and respect chunks with a action of Skip. We don't properly report the values back yet.

### DIFF
--- a/chunk_init_common.go
+++ b/chunk_init_common.go
@@ -49,6 +49,7 @@ type chunkInitCommon struct {
 	numInboundStreams              uint16
 	initialTSN                     uint32
 	params                         []param
+	unrecognizedParams             []paramHeader
 }
 
 const (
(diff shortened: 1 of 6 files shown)


Sean-Der:
If someone has the availability mind looking at this? Soon Chrome is going to roll out with this new param (that we don't support yet)

Pion is going to fail to negotiate SCTP then. I want to get this out so we don't see DataChannels failing.

Sean-Der:
Test failure is ok/expected. We are breaking public API by not having unmarshal errors anymore

chunk_init_common.go
```diff
@@ -91,18 +91,21 @@ func (i *chunkInitCommon) unmarshal(raw []byte) error {
 	remaining := len(raw) - offset
 	for remaining > 0 {
 		if remaining > initOptionalVarHeaderLength {
-			pType, err := parseParamType(raw[offset:])
-			if err != nil {
+			pHeader := paramHeader{}
```
edaniels, line 94:
```suggestion
			var pHeader paramHeader
```

edaniels (approved):
LGTM. Might be helpful to comment on report actions that we don't actually do that and/or point to a gh issue

PR #299, opened 2024-02-22
Reduce RTO.Initial to 1 second

tuexen:
This is the default in [RFC 9260](https://www.rfc-editor.org/rfc/rfc9260.html#section-16).


Sean-Der (approved):

Sean-Der:
Thank you @tuexen !

PR #300, opened 2024-02-23
Fix checksum check for received packets

tuexen:
Only the following packets should be accepted:
* Packets with a correct CRC32c checksum.
* Packets with an incorrect zero checksum, if its acceptance is allowed.

In particular, packets with an incorrect non-zero CRC32c must be discarded.


edaniels:
I'll try to look at this Monday!

tuexen:
> I'll try to look at this Monday!

Great thanks! I also have a local patch to fix the handling of the parameter during the handshake. But I'll submit that one, once this has been dealt with.

edaniels (approved):
Looks good!

PR #301, opened 2024-02-23
Consider padding when bundling chunks in packets

sukunrt:
Currently padding is not considered when decided how many chunks to bundle in to 1 packet. This causes some packets to exceed the MTU size. See the added test for a repro.


edaniels (approved):
Nice!

PR #303, opened 2024-02-27
Configurable RTOMax - fix #181

### LINKED ISSUE #181: Endless buffering of outbound traffic when network is loaded
### Your environment.
  - Version: 1.7.11
  - Browser: iPadOS 14.3

### What did you do?

I developed a terminal server over pion/webrtc, using data channels to pipe input & output. Things work well on a clean network and on a mobile one. I even used netem to emulate high jitter scenario and all works well. Only when I use my home network - a small pipe with two zoom sessions going on in parallel - the terminal stops receiving data.

I'm attaching a log where I have three data channels opened on the same peer. From the logs it looks like the buffer keeps growing and growing and growing

[frozen2.log](https://github.com/pion/sctp/files/5906430/frozen2.log)

daonb:
#### Description

This change adds RTOMax to Config letting the user cap the retransmission timer without breaking compatibility. If RTOMax is 0 the current value of 60 seconds is used.

#### Reference issue
Start fixing #181, needs another PR in webrtc to support this feature

### DIFF
--- a/association.go
+++ b/association.go
@@ -237,6 +237,8 @@ type Config struct {
 	MaxMessageSize       uint32
 	EnableZeroChecksum   bool
 	LoggerFactory        logging.LoggerFactory
+	// RTOMax is the maximum retransmission timeout in milliseconds
+	RTOMax float64
 }
 
 // Server accepts a SCTP stream over a conn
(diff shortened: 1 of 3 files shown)


edaniels:
@daonb have you seen the issue go away if you just keep retransmitting? When testing the issue, I saw no progress happen sometimes even when RTO 3-6 happen

daonb:
@edaniels when congestion clears, the issue goes away and no data is lost. I remember once I had to wait 60 second for a key I hit to appear on the terminal. This will allow time critical low bandwidth data channel based peers to be more aggressive with their retransmission policy and keep the peers synced even in bad internet weather.

edaniels:
gotcha! In that case we should get this in, especially since it's configurable. I have a feeling 181 is exposing more than one issue that goes beyond RTO, but it looks like we don't have any conclusive data yet.

edaniels:
Also @daonb not sure if you're on slack but I started a convo for me you and @enobufs

edaniels (approved):
LGTM

PR #304, opened 2024-02-28
Unmarshal packets only once

tuexen:
Unmarshal of a packet should only be done once, since doing it multiple times ends up in an incorrect list of chunks. Also, perform checksum validation first.

For pion, the only chunks requiring a correct CRC32c are the INIT and COOKIE-ECHO chunks, which need to be the first chunk in a packet.


edaniels (approved):
hehe this is better than my fix here https://github.com/pion/sctp/pull/302/files#diff-b094e403b6f9c6dc300207a21d074085bc52a37548e49aeeb79c17e54f592748R693 :). LGTM

PR #308, opened 2024-02-28
Add SACKs sent and packets stats

edaniels:
#### Description
This helps for debugging and is relatively low cost

### DIFF
--- a/association_stats.go
+++ b/association_stats.go
@@ -8,11 +8,30 @@ import (
 )
 
 type associationStats struct {
-	nDATAs       uint64
-	nSACKs       uint64
-	nT3Timeouts  uint64
-	nAckTimeouts uint64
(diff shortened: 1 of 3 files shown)


association_stats.go
```diff
@@ -8,13 +8,32 @@ import (
 )
 
 type associationStats struct {
+	nPackets     uint64
+	nPacketsSent uint64
```
sukunrt, line 12: NIT: 
I would prefer
``` 
nPacketsReceived
nPacketsSent
```

I understand we would need to change `nSACK` to `nSACKReceived`

sukunrt (approved):

association_stats.go
edaniels: I was avoiding the change but let's do it

PR #309, opened 2024-02-28
Fix zero checksum usage

tuexen:
zero checksum acceptance was implemented as a symmetrical feature, but this is not the way it is specified. Each side announces independently from the peer, that it can accept packets with an incorrect zero checksum. In particular, it is completely valid to send a packet containing an INIT ACK chunk with an incorrect zero checksum, as long as the peer announced the support in the INIT chunk.

### DIFF
--- a/association.go
+++ b/association.go
@@ -177,8 +177,8 @@ type Association struct {
 	cumulativeTSNAckPoint   uint32
 	advancedPeerTSNAckPoint uint32
 	useForwardTSN           bool
-	useZeroChecksum         bool
-	requestZeroChecksum     bool
+	sendZeroChecksum        bool
+	recvZeroChecksum        bool
 
 	// Congestion control parameters
 	maxReceiveBufferSize uint32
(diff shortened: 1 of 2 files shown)


edaniels (approved):
Ah, good catch!

Sean-Der:
@edaniels Can you check if this works with old versions of Pion? Can a v3 of Pion (with this commit) connect to an old version of v3 (like v3.2.15) We might need to disable this by default then if not.

If this now matches the behavior in https://github.com/pion/webrtc/issues/2618

Sean-Der:
Yep! That's unfortunate :/

Thanks for taking care of that

Sean-Der:
Then we can also add to `SettingEngine` for people to opt-in

edaniels:
Never mind, this works fine. I was using my go workspace incorrectly. EnableZeroChecksum is already what causes us to emit the parameter. If that's not enabled, we will never send a parameter that cannot be recognized.

PR #312, opened 2024-03-01
initialize source port and destination port if possible

hugoArregui:


stv0g:
I am wondering, does this break backward compatibility with existing applications which pass a `net.Conn`?

PR #319, opened 2024-04-01
Add receive chunk tracker for better received chunk handling

sukunrt:
This PR introduces a received chunk tracker. 
1. This is optimized for handling incoming chunks and reporting ACK Ranges by doing a sorted insert in received ACK ranges as opposed to sorting the whole list after insert.
2. This removes the reference to the data chunk so that it could be garbage collected quicker.


edaniels:
@sukunrt nice! As you get ready to put this up for review, could you share some performance testing you do/did with this?

JoTurk:
superseded by #337

PR #320, opened 2024-04-01
Optimize payload queue for write path

sukunrt:
Follow up to #319 
This optimizes the payload Queue for the write path by doing sorted inserts to the tsns slice.


JoTurk:
superseded by #337

PR #330, opened 2024-04-05
Fix flaky test by ignoring error in stream close

MarcoPolo:
This test was introduced in #323. Sometimes closing the stream will error with:

> sending reset packet in non-established state: state=Closed

I think it's okay to ignore this error as it's not related to the test.


sukunrt (approved):
We should allow closing a stream twice. It should just be a no op.

MarcoPolo:
Agreed. This will fix the flaky test before we make that change

PR #336, opened 2024-04-16
Reuse time.Timer in ack and rtx timer utilities

paulwe:


edaniels:
@paulwe is this for performance by any chance?

paulwe:
@edaniels yes, we're seeing high `runtime.siftdownTimer` when serving large numbers of clients. these two utilities are responsible for creating over 99% of the `time.Timer` instances in our code.

ack_timer.go
```diff
@@ -41,34 +52,35 @@ func (t *ackTimer) start() bool {
...
+		case <-t.timer.C:
+			t.mutex.Lock()
+			switch t.state {
+			case ackTimerStopped:
+				// if the timer is already stopped empty the cancel channel
+				<-t.cancel
+			case ackTimerStarted:
+				t.state = ackTimerStopped
+				defer t.observer.onAckTimeout()
+			case ackTimerClosed:
```
edaniels, line 74: is it worth emptying t.cancel here as well?

edaniels (approved):
LGTM

ack_timer.go
```diff
@@ -41,34 +52,35 @@ func (t *ackTimer) start() bool {
...
 		return false
 	}
-
-	// this is a noop if the timer is already running
-	if t.stopFunc != nil {
-		return false
-	}
-
-	cancelCh := make(chan struct{})
+	t.state = ackTimerStarted
```
boks1971, line 59: Is there a chance of race here? Is there a need to drain timer channel or t.cancel?

Thinking aloud
1. `stop()` called: that will write to `t.cancel` and set state to `stopped`
2. Assume, the go routine below has not drained `t.cancel` yet.
3. `start()` happens: this will try to restart the timer as state is `stopped`
4. This will launch another go routine. Will this clash with go routine in Step 2?

ack_timer.go
paulwe: state and writes to the cancel channel are synced. 
* if we are in this critical section and the state is `started` we cannot have written to `cancel`. 
* if we change the state from `started` to `stopped` in this critical section then call `stop` afterward the state will be `stopped` and we won't won't write to `cancel`.
* `cancel` could be dirty if the state is `closed` but we won't reuse the timer later so we don't care.

ack_timer.go
paulwe: good catch. invocations of the callback are interchangeable and for this to happen the time between calls has to be practically zero so we can eat the extra cancel and the earlier goroutine will serve as the timeout for the latest call to start.

PR #337, opened 2024-06-27
Improve performance for inflight and payload queue

cnderrauber:
The `payloadQueue.updateSortedKeys` costs more than 40% cpu time when  send sctp packet at high rate with loss/out of order network. The function is frequently called to order the entire map to generate the sorted slice. This change uses individual queue structs for inflight and payload queue. The inflight queue's chunk tsn is always consecutive so use a queue to hold chunks and it is always ordered, and use bitmask to hold payload tsn queue to calculate cumulative tsn and SACK.

Queue benchmark:

```
BenchmarkOldPayloadQueue-10    	     332	   3773029 ns/op	 3416218 B/op	   67217     67217 allocs/op
BenchmarkReceivePayloadQueue-10    	    3188	    347581 ns/op	   19488 B/op      2101 allocs/op
```

Use the sctp send/recv [tester](https://github.com/cnderrauber/sctpbenchmark) in a cpu limited receiver with 50ms rtt network , it also shows lower cpu usage and higher throughput:

```
target data rate 10MB/s, 1024 message size, 50ms rtt
original: receiver 110% cpu usage, sender 50% cpu usage, result 6.36MB/s
changed: receiver 70% cpu usage, sender 30% cpu usage, result 8.75MB/s
```

### DIFF
--- a/receive_payload_queue.go
+++ b/receive_payload_queue.go
@@ -1,0 +1,182 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package sctp
+
+import (
+	"fmt"
+	"math/bits"
+)
+
+type receivePayloadQueue struct {
+	tailTSN      uint32
+	chunkSize    int
+	tsnBitmask   []uint64
+	dupTSN       []uint32
+	maxTSNOffset uint32
+
+	cumulativeTSN uint32
+}
+
+func newReceivePayloadQueue(maxTSNOffset uint32) *receivePayloadQueue {
+	maxTSNOffset = ((maxTSNOffset + 63) / 64) * 64
+	return &receivePayloadQueue{
+		tsnBitmask:   make([]uint64, maxTSNOffset/64),
+		maxTSNOffset: maxTSNOffset,
+	}
+}
+
+func (q *receivePayloadQueue) init(cumulativeTSN uint32) {
+	q.cumulativeTSN = cumulativeTSN
+	q.tailTSN = cumulativeTSN
+	q.chunkSize = 0
+	for i := range q.tsnBitmask {
+		q.tsnBitmask[i] = 0
+	}
+	q.dupTSN = q.dupTSN[:0]
+}
+
+func (q *receivePayloadQueue) hasChunk(tsn uint32) bool {
+	if q.chunkSize == 0 || sna32LTE(tsn, q.cumulativeTSN) || sna32GT(tsn, q.tailTSN) {
+		return false
+	}
+
+	index, offset := int(tsn/64)%len(q.tsnBitmask), tsn%64
+	return q.tsnBitmask[index]&(1<<offset) != 0
+}
+
+func (q *receivePayloadQueue) canPush(tsn uint32) bool {
+	ok := q.hasChunk(tsn)
+	if ok || sna32LTE(tsn, q.cumulativeTSN) || sna32GT(tsn, q.cumulativeTSN+q.maxTSNOffset) {
+		return false
+	}
+	return true
+}
+
+// push pushes a payload data. If the payload data is already in our queue or
+// older than our cumulativeTSN marker, it will be recored as duplications,
+// which can later be retrieved using popDuplicates.
+func (q *receivePayloadQueue) push(tsn uint32) bool {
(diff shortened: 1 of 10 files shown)


edaniels:
fantastic improvement

cnderrauber:
> Amazing! Those numbers are really exciting @cnderrauber
> 
> Would it be possible to drop the dependency?

@Sean-Der  It can be done by implementing a queue/ring buffer. Do we try to avoid all third-party dependency as possible or just concern this new imported dequeue?

Sean-Der:
@cnderrauber I try and avoid all third-party dependencies. Currently we only have 3rd party dependencies for testing. Downstream breaking/doing unexpected things has been a problem

receive_payload_queue.go
```diff
@@ -0,0 +1,201 @@
...
+	dups := q.dupTSN
+	q.dupTSN = []uint32{}
+	return dups
+}
+
+func (q *receivePayloadQueue) getGapAckBlocks() (gapAckBlocks []gapAckBlock) {
+	var b gapAckBlock
+
+	if q.chunkSize == 0 {
+		return []gapAckBlock{}
```
paulwe, line 108: return nil?

receive_payload_queue.go
```diff
@@ -0,0 +1,201 @@
...
+				findEnd = true
+			} else {
+				// no result, find start bits in next uint64 bitmask
+				tsn += uint32(64 - offset)
+			}
+		} else {
+			if zeroBit, ok := getFirstZeroBit(q.tsnBitmask[index], offset, 64); ok {
+				b.end = uint16(tsn + uint32(zeroBit-offset) - 1 - q.cumulativeTSN)
+				tsn += uint32(zeroBit - offset)
+				gapAckBlocks = append(gapAckBlocks, gapAckBlock{
```
paulwe, line 129: if there are more than one gap in the bitmap is it ok that we're only returning the first one?

receive_payload_queue.go
```diff
@@ -0,0 +1,201 @@
...
+
+	if q.chunkSize == 0 {
+		return []gapAckBlock{}
+	}
+
+	startTSN, endTSN := q.cumulativeTSN+1, q.tailTSN
+	var findEnd bool
+	for tsn := startTSN; sna32LTE(tsn, endTSN); {
+		index, offset := int(tsn/64)%len(q.tsnBitmask), int(tsn%64)
+		if !findEnd {
```
paulwe, line 115: should we split this into 2 loops instead of branching in every iteration?

paulwe (approved):
lgtm 🚀

cnderrauber:
> @cnderrauber I try and avoid all third-party dependencies. Currently we only have 3rd party dependencies for testing. Downstream breaking/doing unexpected things has been a problem

Sounds good, will remove it

receive_payload_queue.go
```diff
@@ -0,0 +1,201 @@
...
+		return 0, false
+	}
+
+	for i := start; i < end; i++ {
+		if val&(1<<i) == 0 {
+			return i, true
+		}
+	}
+	return 0, false
+}
```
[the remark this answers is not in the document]
cnderrauber: TIL there is a `bits`, the snippet can't pass test, will look at the package .

receive_payload_queue.go
```diff
@@ -0,0 +1,201 @@
...
+		return false
+	}
+
+	index, offset := int(tsn/64)%len(q.tsnBitmask), tsn%64
+	return q.tsnBitmask[index]&(1<<offset) != 0
+}
+
+func (q *receivePayloadQueue) canPush(tsn uint32) bool {
+	ok := q.hasChunk(tsn)
+	if ok || sna32LTE(tsn, q.cumulativeTSN) || sna32GT(tsn, q.cumulativeTSN+q.maxTSNOffset) {
```
boks1971, line 50: ~~should this be GTE?~~

~~Assuming cumulative is 0 and maxTSN is 64, I am guessing it can hold 0..63 and 64 is out of range?~~

boks1971: oh, LTE, so cumulativeTSN is not in range, GT is okay then.

boks1971 (approved):
lgtm! Awesome stuff @cnderrauber

PR #338, opened 2024-07-01
Fix inconsistencies in timer implementations

paulwe:
Replaces goroutine based ack timer with callback pattern used in other parts of the project.

Adds pending timer tick counters to prevent races between Stop and timers firing. This fixes an edge case where callacks may have fired after stop was called.


cnderrauber (approved):
looks good!

PR #341, opened 2024-07-09
Ensure all INIT have a verification tag of 0 as per the spec

jmelancongen:
#### Description

We've been seeing issues with one of our tests that sometime hangs while opening a DataChannel between two peers. Looking at the detailed logs we've been seeing this pattern:

> [Producer] sending INIT
> [Consumer] failed validating packet init chunk expects a verification tag of 0 on the packet when out-of-the-blue
> [Consumer] sending INIT
> [Producer] failed validating packet init chunk expects a verification tag of 0 on the packet when out-of-the-blue
> ...2 seconds pause...
> [Producer] sending INIT
> [Consumer] failed validating packet init chunk expects a verification tag of 0 on the packet when out-of-the-blue
> [Consumer] sending INIT
> [Producer] failed validating packet init chunk expects a verification tag of 0 on the packet when out-of-the-blue

and the data channel never opens

We've found out that there is a PR from a year ago in webrtc-rs that seems to fit exactly the issue we are hitting:
webrtc-rs/webrtc#405

Where the fix was to ensure the init packet always uses the 0 tag, instead of using the Association state which might be different? We're just bringing that same fix here.


Sean-Der:
@jmelancongen Is this Pion on both sides when you see this happen? I would like to create a reproduce (I can help with that!)

jmelancongen:
Here is the SCTP handshake part of our test: [sctp-handshake-hang.txt](https://github.com/user-attachments/files/16177226/sctp-handshake-hang.txt)

From what I was told, the test runs every minute in multiple locations, and it reproduces the issue once every 2-3 weeks globally. I've boiled down the logs to these bits:

```
[00:10:46.9180-Dbg] [producer] [0xc0002401e0] sending INIT
[00:10:46.9631-Dbg] [consumer] [0xc0002723c0] sending INIT
[00:10:46.9632-Dbg] [consumer] [0xc0002723c0] chunkInit received in state 'CookieWait'
[00:10:46.9632-Dbg] [consumer] sending INIT ACK
[00:10:47.0227-Dbg] [producer] [0xc0002401e0] chunkInit received in state 'CookieWait'
[00:10:47.0227-Dbg] [producer] sending INIT ACK
[00:10:47.9187-Dbg] [producer] [0xc0002401e0] sending INIT
[00:10:47.9491-Wrn] [consumer] [0xc0002723c0] failed validating packet init chunk expects a verification tag of 0 on the packet when out-of-the-blue
[00:10:47.9632-Dbg] [consumer] [0xc0002723c0] sending INIT
[00:10:47.9954-Wrn] [producer] [0xc0002401e0] failed validating packet init chunk expects a verification tag of 0 on the packet when out-of-the-blue
```

I believe the specific problem stems from: 
* High enough latency that both sides send their INIT before receiving the other side
* Both sides receive the INIT of the other
* Both INIT ACK are lost

I'm not too sure how to isolate those conditions in a smaller self-contained reproducer though. Since it feels like it needs the relay to be triggered. I'm also not very familiar with the testing framework for pion sadly, but it might be easier to reproduce this way?

Sean-Der:
Fantastic! I can write a test for that today, that will be pretty easy to reproduce.

jmelancongen:
Sorry for the ping but did you have time to check on that? I'd would love to have the fix so we can remove a workaround on our side

joeturki:
@jmelancongen Hello I was able to make a test for your PR https://github.com/pion/sctp/pull/359 once it's confirmed & reviewed, I'll merge yours :)

joeturki (approved):

jmelancongen:
Awesome that's great news. Thanks a lot for looking at it!

Sean-Der:
I am so sorry @jmelancongen I lost track on this one.

Thank you @JoeTurki you are amazing! So appreciative of you getting into this stuff :) 

I will merge this one, and then the test!

PR #350, opened 2024-08-14
set close state before unregisterStream

edaniels:
Although `OpenStream` checks state before creating a stream, `readLoop`'s `unregisterStream` can be called outside of a direct `Close`. When this happens, `setState(closed)` happens **after** unregisterStream which is an incorrect ordering causing an `OpenStream` to proceed. readLoop should set the closed state prior to unregistering streams.

Repro with carefully placed sleeps to mimic an internal close of connections being closed:
```
diff --git a/association.go b/association.go
index 29e9978..2cf2539 100644
--- a/association.go
+++ b/association.go
@@ -578,6 +578,7 @@ func (a *Association) readLoop() {
 			a.unregisterStream(s, closeErr)
 		}
 		a.lock.Unlock()
+		println("unregged")
 		close(a.acceptCh)
 		close(a.readLoopCloseCh)
 
@@ -650,6 +651,8 @@ loop:
 		}
 	}
 
+	time.Sleep(3 * time.Second)
+	println("now closed")
 	a.setState(closed)
 	a.closeAllTimers()
 }
@@ -1504,13 +1507,18 @@ func (a *Association) getMyReceiverWindowCredit() uint32 {
 
 // OpenStream opens a stream
 func (a *Association) OpenStream(streamIdentifier uint16, defaultPayloadType PayloadProtocolIdentifier) (*Stream, error) {
+	time.Sleep(time.Second)
 	a.lock.Lock()
 	defer a.lock.Unlock()
 
+	println("adding")
+
 	switch a.getState() {
 	case shutdownAckSent, shutdownPending, shutdownReceived, shutdownSent, closed:
+		println("no")
 		return nil, ErrAssociationClosed
 	}
+	println("yes")
 
 	return a.getOrCreateStream(streamIdentifier, false, defaultPayloadType), nil
 }
```

### DIFF
--- a/association.go
+++ b/association.go
@@ -573,6 +573,7 @@ func (a *Association) readLoop() {
 		a.closeWriteLoopOnce.Do(func() { close(a.closeWriteLoopCh) })
 
 		a.lock.Lock()
+		a.setState(closed)
 		for _, s := range a.streams {
 			a.unregisterStream(s, closeErr)
 		}
--- a/association_test.go
+++ b/association_test.go
@@ -3475,3 +3475,25 @@ func TestAssociation_OpenStreamAfterClose(t *testing.T) {
 	_, err = a2.OpenStream(1, PayloadTypeWebRTCString)
 	require.ErrorIs(t, err, ErrAssociationClosed)
 }
+
+// https://github.com/pion/sctp/pull/350
+// may need to run with a high test count to reproduce if there
+// is ever a regression.
+func TestAssociation_OpenStreamAfterInternalClose(t *testing.T) {
+	checkGoroutineLeaks(t)
+
+	a1, a2, err := createAssocs()
+	require.NoError(t, err)
+
+	a1.netConn.Close()
+	a2.netConn.Close()
+
+	a1.OpenStream(1, PayloadTypeWebRTCString)
+	a2.OpenStream(1, PayloadTypeWebRTCString)
+
+	require.NoError(t, a1.Close())
+	require.NoError(t, a2.Close())
+
+	require.Equal(t, 0, len(a1.streams))
+	require.Equal(t, 0, len(a2.streams))
+}


sukunrt:
>readLoop's unregisterStream can be called outside of a direct Close

When does this happen?

edaniels:
> > readLoop's unregisterStream can be called outside of a direct Close
> 
> When does this happen?

In the particular real case I noticed, a DTLS CloseNotify was received which closes the DTLS Conn decryption channel which sets all Reads to be EOFs. That is then received by the SCTP Assocation's readLoop and it unrolls from there.

sukunrt:
I think this change is fine. I believe the case you're mentioning happens when we close a webrtc connection and pion/webrtc closes the peerconnection by closing the DTLS connection and doesn't send an abort as the specs recommend: 

> 6. If connection.[[[SctpTransport]]](https://www.w3.org/TR/webrtc/#dfn-sctptransport) is not null, tear down the underlying SCTP association by sending an SCTP ABORT chunk and set the [[[SctpTransportState]]](https://www.w3.org/TR/webrtc/#dfn-sctptransportstate) to "[closed](https://www.w3.org/TR/webrtc/#idl-def-RTCSctpTransportState.closed)"

https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-close

sukunrt (approved):

edaniels:
> I think this change is fine. I believe the case you're mentioning happens when we close a webrtc connection and pion/webrtc closes the peerconnection by closing the DTLS connection and doesn't send a SCTP Abort chunk as the specs recommend:
> 
> > 6. If connection.[[[SctpTransport]]](https://www.w3.org/TR/webrtc/#dfn-sctptransport) is not null, tear down the underlying SCTP association by sending an SCTP ABORT chunk and set the [[[SctpTransportState]]](https://www.w3.org/TR/webrtc/#dfn-sctptransportstate) to "[closed](https://www.w3.org/TR/webrtc/#idl-def-RTCSctpTransportState.closed)"
> 
> https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-close

Yes I think that's the case but in the one I'm noticing the other side is shutting down and sends the DTLS message. It probably also sends the SCTP abort but maybe there's a race there.

edaniels:
> I think this change is fine. I believe the case you're mentioning happens when we close a webrtc connection and pion/webrtc closes the peerconnection by closing the DTLS connection and doesn't send a SCTP Abort chunk as the specs recommend:
> 
> > 6. If connection.[[[SctpTransport]]](https://www.w3.org/TR/webrtc/#dfn-sctptransport) is not null, tear down the underlying SCTP association by sending an SCTP ABORT chunk and set the [[[SctpTransportState]]](https://www.w3.org/TR/webrtc/#dfn-sctptransportstate) to "[closed](https://www.w3.org/TR/webrtc/#idl-def-RTCSctpTransportState.closed)"
> 
> https://www.w3.org/TR/webrtc/#dom-rtcpeerconnection-close

Wrote my reply too soon :). Yes, you're correct. I think due to DTLS being unordered, there's a chance we can still get that notification even if the SCTP abort is sent. There's also the case of bad implementations cause this race to happen as well.

PR #353, opened 2024-11-14
Increment missing indications from SACK

cnderrauber:
According to the [RFC4960](https://datatracker.ietf.org/doc/html/rfc4960#section-7.2.4) , the missing indications should be incremented for missing chunks in the SACK.

> If an endpoint is in Fast
>    Recovery and a SACK arrives that advances the Cumulative TSN Ack
>    Point, the miss indications are incremented for all TSNs reported
>    missing in the SACK.

### DIFF
--- a/association.go
+++ b/association.go
@@ -1755,7 +1755,7 @@ func (a *Association) onCumulativeTSNAckPointAdvanced(totalBytesAcked int) {
 }
 
 // The caller should hold the lock.
-func (a *Association) processFastRetransmission(cumTSNAckPoint, htna uint32, cumTSNAckPointAdvanced bool) error {
+func (a *Association) processFastRetransmission(cumTSNAckPoint uint32, gapAckBlocks []gapAckBlock, htna uint32, cumTSNAckPointAdvanced bool) error {
 	// HTNA algorithm - RFC 4960 Sec 7.2.4
 	// Increment missIndicator of each chunks that the SACK reported missing
 	// when either of the following is met:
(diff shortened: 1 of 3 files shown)


association.go
```diff
@@ -1855,7 +1858,7 @@ func (a *Association) handleSack(d *chunkSelectiveAck) error {
 		a.setRWND(d.advertisedReceiverWindowCredit - bytesOutstanding)
 	}
 
-	err = a.processFastRetransmission(d.cumulativeTSNAck, htna, cumTSNAckPointAdvanced)
+	err = a.processFastRetransmission(d.cumulativeTSNAck, d.gapAckBlocks, htna, cumTSNAckPointAdvanced)
```
boks1971, line 1861: Don't know the protocol, but just reading this, can this just pass in the `end` of the last entry? Any reason for passing in the slice?

boks1971: Guess, you will have to do the check for length here rather than inside the function. Maybe, six of one and half-a-dozen of other case?

cnderrauber: No difference but I prefer to put the code and comment together to understand `all TSNs reported missing from the SACK (gapAckBlocks)`

boks1971: Thank you @cnderrauber

boks1971 (approved):
LGTM, but I do not know the protocol to comment on nuances of this. Would be good to get another pair of 👀 on this.

Sean-Der:
Mind adding a test @cnderrauber! It’s been a while going to re-read this section

PR #354, opened 2024-11-18
Add congestion control parameters to config

cnderrauber:
The loss-based congestion control get poor
performance under high bandwidth, high rtt
and packet loss case since the congestion
window becomes 1 mtu and increase slowly
after retransmit timeout. And fast recovery
retransmit cause exit slowly in consecutive
packet loss. This change add paramters to
the config then the user can set them to get
higher throughput in such cases.

#### Reference issue
Fixes #...

### DIFF
--- a/association.go
+++ b/association.go
@@ -212,6 +212,9 @@ type Association struct {
 	partialBytesAcked    uint32
 	inFastRecovery       bool
 	fastRecoverExitPoint uint32
+	minCwnd              uint32 // Minimum congestion window
+	fastRtxWnd           uint32 // Send window for fast retransmit
+	cwndCAStep           uint32 // Step of congestion window increase at Congestion Avoidance
 
 	// RTX & Ack timer
 	rtoMgr     *rtoManager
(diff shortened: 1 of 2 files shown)


boks1971 (approved):
lgtm, is it possible to add a UT for this?

cnderrauber:
Figure_1

The benchmark shows improvements under high rtt/loss condition with custom congestion parameters, 50mpbs@200ms, 6.48mbps@100ms/1%loss, 3.76mbps@200ms/1%loss, compared to original 37mbps@200ms, 1.36mbps@100ms/1%loss, 0.72mbps@200ms/1%loss.

Sean-Der:
@cnderrauber fantastic!

I don't fully understand the change. Would it be possible to make the diff only be related to congestion control? Like the `bundleDataChunksIntoPackets`

Sean-Der:
What does usrsctp do here? @tuexen Do you have any opinion/any insights? What should we be doing better

cnderrauber:
> what does control look like? it would be great to see the before & after benchmarks.

fastrtx
Here is 2 linux server with `tc` to simulate 100ms rtt, with default parameter the fast recovery costs seconds to exit and the congestion window is throlttled, and a large fast retransmit window make it exit quickly and the congestion window increase quickly as well as the bitrates.

cnderrauber:
> @cnderrauber fantastic!
> 
> I don't fully understand the change. Would it be possible to make the diff only be related to congestion control? Like the `bundleDataChunksIntoPackets`

All changes and paramters are related to the congestion control, what does ` only be related to congestion control` mean?

Issue #314, opened 2024-03-06
v2 Plans


edaniels:
@enobufs, I figured we could start talking here publicly about what we should do in v2.

Right now we know that ultimately we want:
- RACK
- TLP
- Blocking writes

The first two make sense to me and can likely be done without an official v2. Blocking writes however is a breaking change and I was curious if you had some ideas in your head on what a newer architecture would look like? I know you mentioned that you don't love the current state machine.

I figure we can start here and try to outline the general idea of what we should do. I don't think the plan has to be hyper specific but it would be good to be on the same page and be able to define units of work that don't result in just some massive PR to get to v2 (unless that is truly the best way!).

edaniels:
Also, for what it's worth, I don't think we're in any rush here, so no pressure to respond ASAP to this.

edaniels:
That's fine with me. Do you have any vision for what you'd re-architect to achieve blocking writes? I could "just" go down a proof of concept route to achieve and see what changes fall out. What do you think?

enobufs:
I see (or hope) not much `re-architing` required here. sctp.Stream is already looks like net.Conn. We have two goroutines: writeLoop and readLoop, which wouldn't change. Read operation is already blocking, so it is Stream.WriteSCTP() and underlying call into Association needs to be modified. I'd say we can just use Association's mutex to block the write operation (Association.sendPayloadData()). Challenges I can see from top of my head is because we have many streams competing over the shared Association, we'd need to figure out how to control the deadlines (e.g. SetDeadline, SetWriteDeadline)

This is my quick thought (I may be wrong). I think it is good idea for you to just go ahead try it out ;)

edaniels:
I'll give it a go once some of my paid work lets up :D

sirzooro:
Please also add fix for https://github.com/pion/sctp/issues/239. It is useful for apps which sends lots of data via DataChannel, and handle network contentions by closing old data channel and opening new one. Now unsent/unacked chunks piles up in inflight queue, increasing memory and CPU usage what is not desirable.

erik-moqvist:
@edaniels @enobufs Would be awesome if RACK is implemented soon.  =)

I've done some SCTP testing over networks where packets are not received in the same order as they are transmitted, and the throughput is generally quite bad when that happens. As I understand it RACK, a time based algorithm, generally outperforms a counting nacks algorithm, in networks with frequent packet reordering.

PR #356, opened 2024-11-26
Add blocking write mode for association

cnderrauber:
Block write is useful for cases like sfu to forward reliable message between clients and detach the data channel to use it as normal golang connections. The default behavior is still non-block mode and the user need to enable block mode by the config. Will also add an SettingEngine option to pion/webrtc for this.

Relates to #77  and #314

### DIFF
--- a/association.go
+++ b/association.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/pion/logging"
 	"github.com/pion/randutil"
+	"github.com/pion/transport/v3/deadline"
 )
 
 // Port 5000 shows up in examples for SDPs used by WebRTC. Since this implementation
@@ -251,6 +252,10 @@ type Association struct {
 	delayedAckTriggered   bool
 	immediateAckTriggered bool
 
+	blockWrite   bool
+	writePending bool
+	writeNotify  chan struct{}
+
 	name string
 	log  logging.LeveledLogger
 }
@@ -264,6 +269,7 @@ type Config struct {
 	MaxMessageSize       uint32
 	EnableZeroChecksum   bool
 	LoggerFactory        logging.LoggerFactory
+	BlockWrite           bool
 
 	// congestion control configuration
 	// RTOMax is the maximum retransmission timeout in milliseconds
@@ -375,6 +381,8 @@ func createAssociation(config Config) *Association {
 		stats:                   &associationStats{},
 		log:                     config.LoggerFactory.NewLogger("sctp"),
 		name:                    config.Name,
+		blockWrite:              config.BlockWrite,
+		writeNotify:             make(chan struct{}, 1),
 	}
 
 	if a.name == "" {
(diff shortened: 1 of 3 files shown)


Sean-Der:
Love this feature! Great idea putting it behind a SettingEngine option. 

Would it be possible to implement this by watching/peeking at `a.pendingQueue.getNumBytes() + a.inflightQueue.getNumBytes()`? Add some kind of notifier/callback that gets fired when it goes to zero.

It would allow us to keep the internals a lot simpler.

cnderrauber:
> Love this feature! Great idea putting it behind a SettingEngine option.
> 
> Would it be possible to implement this by watching/peeking at `a.pendingQueue.getNumBytes() + a.inflightQueue.getNumBytes()`? Add some kind of notifier/callback that gets fired when it goes to zero.
> 
> It would allow us to keep the internals a lot simpler.

It will reduce the throughput if we watch the pending/inflight bytes to become zero that every write needs to wait for previous one to be acked by remote peer, it is also not identical to the net.conn's write method which returns after copy the data to the socket buffer.

association.go
```diff
@@ -2375,21 +2403,35 @@ func (a *Association) bundleDataChunksIntoPackets(chunks []*chunkPayloadData) []
...
 		return fmt.Errorf("%w: state=%s", ErrPayloadDataStateNotExist,
 			getAssociationStateString(state))
 	}
 
+	if a.blockWrite {
+		for a.writePending {
+			a.lock.Unlock()
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
```
boks1971, line 2421: how should this error be handled? does this bubble up to the app level or it is all handled internally? Am curious about how this affects reliable transmission mode and if app need to retry send?

boks1971 (approved):
lgtm!

association.go
cnderrauber: It will return to upper application, when `WriteSCTP` returns error, it means the data has not been sent (incorrect state or deadline timeout) so no retransmission here, the retransmission only works for sent messages (write operation succeeded).

boks1971: got it, wondering if the application is supposed to try `Write` again?

cnderrauber: It should do that if blocking mode is enabled, like other connections. The default behavior is still unblock mode

Sean-Der:
@cnderrauber @boks1971 Push back if you disagree, but I think we should have `Write` drive and not create these single purpose points of communication.

This would be a deeper change. I think it will be easier to maintain/follow. As more gets added to the code base it gets harder to reason about and eventually the complexity makes everyone want to start over.

cnderrauber:
@Sean-Der It's a patch for blocking write based on currently read/write loop, the feature is useful for flow control in sfu forwarding reliable message and other people request it for long time. I think the change is easy to understand that the queueBuffer works as a ongoing write packet buffer in block mode and other write needs to wait for it drained.

I agree the write drive will make the write more straghtforward but we might still need buffer/notify to implement non-block writing if we don't want to break the API. So it is more like a v2 feature, what do you think?

PR #364, opened 2025-02-18
Set rwnd from InitChunk

cnderrauber:
The rwnd has not been initialized from InitChunk
then payload data will be hold in the pending
queue until first SACK is received.


boks1971 (approved):
lgtm!

PR #365, opened 2025-03-04
Don't discard data on io.ErrShortBuffer

Sean-Der:
Currently if the user passes a buffer to read into that is shorter then the current message we lose the buffer. This PR changes the behavior to instead keep what's in the reassemblyQueue until it can be read.

Relates to pion/webrtc#2712

### DIFF
--- a/association_test.go
+++ b/association_test.go
@@ -165,7 +165,7 @@ func check(err error) {
 func pipeDump() (net.Conn, net.Conn) {
 	aConn := acceptDumbConn()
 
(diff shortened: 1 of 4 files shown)


Sean-Der:
This has a behavior change. Before if a remote sent a large buffer we would discard them. Now we hold on to them, and inform the user they need to call the API again with a larger buffer. 

This is a behavior change, but the behavior before leads to data loss.

JoeTurki (approved):
makes sense :)

PR #369, opened 2025-04-12
Make Association.BufferedAmount public

sirzooro:
It simplifies tracking amount of unsent data for all data channels, including ones in closing state. Especially useful for IoT apps which use multiple data channels and have to monitor memory usage.

I will create another PR in `pion/webrtc` to add `SCTPTransfer.BufferedAmount` there.


edaniels (approved):
LGTM!

PR #392, opened 2025-10-07
VNet remove deprecated rand.read(), improve test readability

philipch07:
#### Description
Removes a deprecated call to `rand.Read()` from the math module in favor of `rand.Read()` from crypto. Also uses testify/require to make some of the `if !assert... { return }` blocks more intuitive/easier to read.

#### Reference issue
N/A


JoeTurki (approved):
ty

PR #397, opened 2025-10-13
Remove zero-checksum heuristic

### LINKED ISSUE #391: Bad ZeroChecksum heuristics
Do we still need heuristics for bad ZeroChecksum implementation? I had case when `a.sendZeroChecksum` flag was incorrectly toggled after few consecutive packets were lost. Could we remove it completely, or disable it by default?

drshrey:
#### Description

Remove the legacy interop heuristic that toggled sendZeroChecksum after repeated RTOs with no TSN progress. Confirmed legacy interop is no longer needed and the flip ends up prolonging loss with newer version peers that don't have the bad zero checksum interpretation logic (where it would assume Zero Checksum Acceptable parameter meant it would receive zero checksum data).

#### Reference issue
https://github.com/pion/sctp/issues/391

Fixes #391

### DIFF
--- a/association.go
+++ b/association.go
@@ -2766,12 +2766,6 @@ func (a *Association) onRetransmissionTimeout(id int, nRtos uint) { //nolint:cyc
 	a.lock.Lock()
 	defer a.lock.Unlock()
 
-	// TSN hasn't been incremented in 3 attempts. Speculatively
-	// toggle ZeroChecksum because old Pion versions had a broken implementation
-	if a.cumulativeTSNAckPoint+1 == a.initialTSN && nRtos%3 == 0 {
-		a.sendZeroChecksum = !a.sendZeroChecksum
-	}
-
 	if id == timerT1Init {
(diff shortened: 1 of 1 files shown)


association.go
```diff
@@ -270,6 +271,12 @@ type Config struct {
 	MaxReceiveBufferSize uint32
 	MaxMessageSize       uint32
 	EnableZeroChecksum   bool
+	// EnableLegacyZeroChecksumInterop enables a legacy interop fallback: when an
+	// association experiences repeated retransmission timeouts without TSN progress,
+	// alternate between sending DATA with and without checksums. Default is false;
+	// enable only when interoperating with legacy peers known to mis-handle
+	// ZeroChecksum negotiation.
+	EnableLegacyZeroChecksumInterop bool
```
JoTurk: what is the purpose of this flag and how are we going to use it from webrtc? can we just remove it all together?

PR #398, opened 2025-10-13
Retract v1.8.12 due to ZeroChecksum misinterpretation

drshrey:
#### Reference issue
Fixes #...

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,3 +16,6 @@ require (
 )
 
 go 1.21
+
+// Retract version with ZeroChecksum misinterpretation (bi-directional/global handling)
+retract v1.8.12


go.mod
```diff
@@ -16,3 +16,7 @@ require (
 )
 
 go 1.21
+
+// Retract versions that included the RTO-based ZeroChecksum flip heuristic
+// which could prolong packet loss
+retract [v1.8.14, v1.8.39]
```
JoTurk: We mainly need to retract v1.8.12. that introduced the bad zerochecksum implementation.

Thinking about it about i'm hesitant to retract such a large range of versions unless the issue is severe, especially since we upgraded the Go version recently, which will make this retraction irrelevant for most of our users because they can't upgrade easily. the flip bug wasn't a big issue!

what do you think?

JoTurk: I looked and it was fixed in v1.8.13 so it's only one version

drshrey: Yeah that makes a lot of sense. These versions aren't necessarily broken, they just prolong loss recovery which isn't enough to justify a retraction and so a retraction is probably overkill. `v1.8.12` is severe because it misinterprets the `ZeroChecksum Acceptable` param as bi-directional. Will reflect to just be that version then!

PR #264, opened 2023-02-16
Fix pendingBaseQueue memory

yann-y:
#### Description
The pendingBaseQueue array is converted to a list.
When transferring large files, the array will not be gc, which takes up a lot of memory space.

PR #401, opened 2025-10-17
Fix a leak with pop in pending_queue

philipch07:
#### Description
A leak was found in pop where `pendingBaseQueue.pop()` would have the slice keep a reference to the popped element.

I also added a regression test to prevent us from having this issue again.

I am unsure if we still want to turn this into a list to allow for gc as referenced in here: https://github.com/pion/sctp/pull/264 considering that we still want the v2 rewrite. I don't mind looking into rebasing that PR if there's any demand for it.

#### Reference issue
The issue was reported here: https://github.com/pion/sctp/pull/264#issuecomment-2606438097

### DIFF
--- a/pending_queue.go
+++ b/pending_queue.go
@@ -25,8 +25,15 @@ func (q *pendingBaseQueue) pop() *chunkPayloadData {
 	if len(q.queue) == 0 {
 		return nil
 	}
+
 	c := q.queue[0]
-	q.queue = q.queue[1:]
+	q.queue[0] = nil
+
+	if len(q.queue) == 0 {
+		q.queue = nil
+	} else {
+		q.queue = q.queue[1:]
+	}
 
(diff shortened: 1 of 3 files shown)


pending_queue.go
```diff
@@ -25,7 +25,9 @@ func (q *pendingBaseQueue) pop() *chunkPayloadData {
 	if len(q.queue) == 0 {
 		return nil
 	}
+
 	c := q.queue[0]
+	q.queue[0] = nil
 	q.queue = q.queue[1:]
```
JoTurk, line 37: nit: we can improve memory usage here if we drop q.queue by setting it to nil (drop capacity), if len is zero.

JoeTurki (approved):
Thank you, also cool that you fixed the typo and thank you for the test.

Issue #402, opened 2025-10-17
Update to RFC 9260


philipch07:
Currently, the SCTP implementation is based off of [RFC 4960](https://www.rfc-editor.org/rfc/rfc4960.html) which is obseleted by [RFC 9260](https://www.rfc-editor.org/rfc/rfc9260.html). It would be nice to update SCTP to RFC 9260!

This issue can help to keep track of the different things that were necessary to update to RFC 9260. I'm not super sure about creating sub-issues for everything, but if anyone wants to help out, it would probably be a good way to manage tasks!

Issue #124, opened 2020-06-16
Add fuzz testing


Sean-Der:
https://github.com/pion/dtls/blob/master/fuzz.go is a good place to start

PR #403, opened 2025-10-17
Update the rtx_timer to RFC 9260

philipch07:
#### Description
- Updates the rtx_timer to [RFC 9260](https://www.rfc-editor.org/rfc/rfc9260.html).
- Fixes typos
- Adds unit tests + fuzz tests (#124)

#### Reference issue
A part of #402.


JoTurk (changes requested):
We'll have to merge this with other PRs and finish since the rtx_timer added here isn't used.

PR #406, opened 2025-10-18
Upgrade Zero Checksum Draft Implementation to RFC 9653

philipch07:
#### Description
- Adds support for [RFC 9653](https://www.rfc-editor.org/rfc/rfc9653).
- Adds fuzz tests to validate the RFC 9653 core properties.

~~Requires #407 to be merged first. Merged!~~ This got reverted.

#442 reintroduces it, so this must be merged first.

#### Reference issue
A part of #124. Closes #405.

PR #407, opened 2025-10-19
Update packet to RFC 9260

philipch07:
#### Description
- Updates packet.go to conform to [RFC 9260](https://www.rfc-editor.org/rfc/rfc9260).
- Adds fuzz tests for packet un/marshalling behavior.

Note: The chunk test is modified since it uses packets, which has changed.

#### Reference issue
A part of #402. Split from #406.

### DIFF
--- a/packet.go
+++ b/packet.go
@@ -18,7 +18,7 @@ var castagnoliTable = crc32.MakeTable(crc32.Castagnoli) // nolint:gochecknogloba
 var fourZeroes [4]byte // nolint:gochecknoglobals
 
 /*
-Packet represents an SCTP packet, defined in https://tools.ietf.org/html/rfc4960#section-3
+Packet represents an SCTP packet, defined in https://tools.ietf.org/html/rfc9260#section-3
 An SCTP packet is composed of a common header and chunks.  A chunk
 contains either control information or user data.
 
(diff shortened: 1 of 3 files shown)


packet.go
```diff
@@ -171,13 +193,22 @@ func (p *packet) marshal(doChecksum bool) ([]byte, error) {
...
 
+	// RFC 9653 section 5.2: send a correct CRC32c when the packet contains INIT or
+	// COOKIE ECHO. Otherwise, if the caller indicates ZCA is active for this
+	// path (doChecksum==true), emit zero.
+	// doChecksum here means zero is allowed for this path.
+	if doChecksum {
+		for _, c := range p.chunks {
+			switch c.(type) {
+			case *chunkInit, *chunkCookieEcho:
+				doChecksum = false
```
JoTurk: not sure about having state / side effect in marshal is this necessary to do here?

philipch07: Thanks for catching this, I was messing with some variable names and accidentally left this in. I changed it to use a helper so it's a little more straightforward (and it avoids manipulating the input).

JoeTurki (approved):
sorry i was supposed to do approve + nitpick not just comment :)

PR #411, opened 2025-10-30
Update stream to RFC 9260

### LINKED ISSUE #410: Update stream.go to RFC 9260
See title.

Nothing really needs to be changed here from what I can tell, but direct citations have been made for RFC 3758 and RFC 7496 for better documentation.

philipch07:
#### Description
Updates stream.go to RFC 9260. Direct citations have been made for RFC 3758 and RFC 7496 for better documentation.

#### Reference issue
Resolves #410.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -106,9 +106,21 @@ func (s *Stream) SetReliabilityParams(unordered bool, relType byte, relVal uint3
 func (s *Stream) setReliabilityParams(unordered bool, relType byte, relVal uint32) {
 	s.log.Debugf("[%s] reliability params: ordered=%v type=%d value=%d",
 		s.name, !unordered, relType, relVal)
+
 	s.unordered = unordered
-	s.reliabilityType = relType
-	s.reliabilityValue = relVal
+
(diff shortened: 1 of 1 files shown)


stream.go
```diff
@@ -106,9 +106,21 @@ func (s *Stream) SetReliabilityParams(unordered bool, relType byte, relVal uint3
...
+	if relType == ReliabilityTypeRexmit || relType == ReliabilityTypeTimed {
+		s.reliabilityType = relType
+		s.reliabilityValue = relVal
+
+		return
+	}
+
+	// for ReliabilityTypeReliable and unknown types, fall back to fully reliable
+	s.reliabilityType = ReliabilityTypeReliable
+	s.reliabilityValue = 0
```
JoTurk, line 123: I don't think this is required by the spec we shouldn't normalize to ReliabilityTypeReliable, I'm going to drop this.

JoTurk:
I don't think this change is correct or required by the spec we should not normalize unknown relType to reliable, and if we remove this, this PR becomes no-op.

PR #425, opened 2025-10-31
Update error_cause to RFC 9260

### LINKED ISSUE #424: Update error_cause files to RFC 9260
Each of these files generally lack a reference to RFC 9260 and they don't check for length first before returning the error cause.

philipch07:
#### Description
- Updated `error_cause.go` to have updated errors according to RFC 9260
- Added an error for other causes if the length is too short
- Added references to RFC 9260 where applicable

#### Reference issue
Resolves #424.

### DIFF
--- a/error_cause.go
+++ b/error_cause.go
@@ -9,7 +9,7 @@ import (
 	"fmt"
 )
 
-// errorCauseCode is a cause code that appears in either a ERROR or ABORT chunk.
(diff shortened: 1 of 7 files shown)


JoTurk (changes requested):
looks good, but need to drop most of the tests, I'll do that tonight.

PR #433, opened 2025-11-08
Update remaining chunk files to RFC 9260

### LINKED ISSUE #426: Update chunk stuff to RFC 9260
There's a lot of chunk stuff

philipch07:
#### Description
Update the remaining chunk files to RFC 9260 with the exception of chunk-reconfig which follows RFC 6525 and requires more updates in association.go.

Requires #427 to be merged first.

#### Reference issue
Resolves #426.


JoeTurki:
i think we should break this into multiple PRs, I'm fine with reviewing it as a single one tho.

PR #438, opened 2025-11-24
Add rack active RTT measurement

philipch07:
#### Description
According to https://duepublico2.uni-due.de/servlets/MCRFileNodeServlet/duepublico_derivate_00073893/Diss_Weinrank.pdf#page=113 section 7.3.2, an enhancement for RACK in SCTP is to add active RTT measurement.

This branch adds that feature. Note that it's built on the RACK branch merged with the adaptive burst mitigation branch.

The best thing to add after this would be the I bit as mentioned in 7.3.4 which is formalized in RFC 9260. That's something that will take place after all the 9260 prs go in, as it involves updating `Association.go` which really needs a refactor since it's way too unwieldy...

#### Reference issue
A RACK enhancement!

### DIFF
--- a/association.go
+++ b/association.go
@@ -1960,14 +1960,14 @@ func (a *Association) processSelectiveAck(selectiveAckChunk *chunkSelectiveAck) 
 					a.log.Tracef("[%s] SACK: measured-rtt=%f srtt=%f new-rto=%f",
 						a.name, rtt, srtt, a.rtoMgr.getRTO())
 				}
+			}
 
-				// RFC 8985 (RACK) sec 5.2: RACK.segment is the most recently sent
-				// segment that has been delivered, including retransmissions.
-				if chunkPayload.since.After(newestDeliveredSendTime) {
-					newestDeliveredSendTime = chunkPayload.since
-					newestDeliveredOrigTSN = chunkPayload.tsn
-					deliveredFound = true
-				}
+			// RFC 8985 (RACK) sec 5.2: RACK.segment is the most recently sent
+			// segment that has been delivered, including retransmissions.
+			if chunkPayload.since.After(newestDeliveredSendTime) {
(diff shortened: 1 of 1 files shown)


association.go
```diff
@@ -1714,34 +1925,54 @@ func (a *Association) processSelectiveAck(selectiveAckChunk *chunkSelectiveAck)
...
+				}
+
+				// RFC 8985 (RACK) sec 5.2: RACK.segment is the most recently sent
+				// segment that has been delivered, including retransmissions.
+				if chunkPayload.since.After(newestDeliveredSendTime) {
+					newestDeliveredSendTime = chunkPayload.since
+					newestDeliveredOrigTSN = chunkPayload.tsn
+					deliveredFound = true
+				}
 			}
```
JoTurk: I found through SCP that this cases regression in RTT tests with dynamic re-order and 3% dropped packets, after looking in pprof and some debugging, I think this was a typo.
> In reality, as an algorithm, RACK does not arm a timer for every segment sent because it's not necessary. Instead, the sender records the most recent transmission time of every data segment sent, including retransmissions. For each ACK received, the sender calculates the latest RTT measurement (if eligible) and adjusts the expiration time of every segment sent but not yet delivered. If a segment has expired, RACK marks it as lost.
> https://datatracker.ietf.org/doc/html/rfc8985

```suggestion
			}

			// RFC 8985 (RACK) sec 5.2: RACK.segment is the most recently sent
			// segment that has been delivered, including retransmissions.
			if chunkPayload.since.After(newestDeliveredSendTime) {
				newestDeliveredSendTime = chunkPayload.since
				newestDeliveredOrigTSN = chunkPayload.tsn
				deliveredFound = true
			}
```

philipch07: Thanks for catching this, I fixed it in both places 😅

PR #443, opened 2025-12-20
Stream Schedulers and User Message Interleaving (RFC 8260)

### LINKED ISSUE #435: Implement RFC 8260
https://datatracker.ietf.org/doc/rfc8260/ seems like something that would be really nice to have since it's basically meant for SCTP used as a part of WebRTC!

JoTurk:
#### Description
https://www.rfc-editor.org/rfc/rfc8260.html

resolves https://github.com/pion/sctp/issues/435

This includes customizable stream scheduler, fair WFQ as the default for webrtc use, but users can implement custom ones, or use one of the default ones we provide, round-robin or WFQ with custom weight. or fully custom schedulers that even violates RFC 8260.


JoTurk:
I'm going to get this ready for merge by this weekend!

JoTurk:
Added tests, and bunch of refactors, the PR is more ready for merge now, I'm going to validate it against other implementations through the week and fix any correctness issue, Then I'll mark it as ready for review!

association.go
```diff
@@ -333,12 +340,12 @@ type snapConfig struct {
...
-	LoggerFactory logging.LoggerFactory
-	Name          string
-	NetConn       net.Conn
-
+	LoggerFactory      logging.LoggerFactory
+	Name               string
+	NetConn            net.Conn
 	BlockWrite         bool
 	EnableZeroChecksum bool
+	EnableInterleaving bool
```
JoTurk, line 348: This PR was made before we moved SCTP to options patterns, Should make it options only.

JoTurk:
(One issue with this PR is the lack of a clear reference implementations to test against, there are few but most implementations feel after thought, or super abstracted).

ValorZard:
@JoTurk this one seems to have pretty good support for this extension:
https://github.com/webrtc/dcsctp

JoTurk:
@ValorZard Thank you a lot, this is what i have been using

ValorZard:
@JoTurk 
based on my research (i.e scanning the github repo) https://github.com/sctplab/usrsctp and by extension https://github.com/paullouisageneau/libdatachannel should also support IDATA/8260

JoTurk:
I'm working on stream schedulers, I'll expose an option to pick a custom scheduler or make one, maybe weighted fair queueing should be the default instead of round robin, I'll expose an option to switch or customize.

association_options.go
```diff
@@ -84,6 +84,17 @@ func WithEnableZeroChecksum(b bool) AssociationOption {
...
+// By default this is true.
+func WithEnableInterleaving(b bool) AssociationOption {
+	return sharedOption(func(c *Config) error {
+		c.enableInterleaving = b
+		c.enableInterleavingSet = true
+
+		return nil
+	})
+}
+
```
JoTurk, line 97: Thinking about this, maybe this should be like rack with options to configure interleaving, e.g using a custom stream scheduler or one of the default ones?

ValorZard: Is there any way to “negotiate” interleaving?

JoTurk: yes, and it's implemented, check this commit,  i added e2e tests, interleaving is only enabled if it's enabled at both sides.


 https://github.com/pion/sctp/pull/443/commits/4dd8a01ca5af0a25ea26d16bae263f8fef687063

JoTurk:
> I'm working on stream schedulers, I'll expose an option to pick a custom scheduler or make one, maybe weighted fair queueing should be the default instead of round robin, I'll expose an option to switch or customize.

done, refactored stream scheduler so it's a customize-able interface, with WFQ being the default for webrtc.

JoTurk:
I think the only missing part is how it interacts with SNAP?

JoTurk:
> I think the only missing part is how it interacts with SNAP?

Should be good to go, time for interops.

association_options.go
ValorZard: in that case i think interleaving should be on by default, and you can turn it off if you want for whatever reason

JoTurk: it's on by default currently (if set is false we flip enabled to true on the constructor), we might change that so it's on by default by pion/webrtc 4.3 instead of now :)

ValorZard: makes sense

JoTurk:
works with both firefox and chrome impls, and i tested a bit with few libraries, and it seems to work, I'll updated pion@scp so it has interop tests just before we enable this by default (pion@4.3).

stream_test.go
```diff
@@ -70,3 +98,209 @@ func TestSessionBufferedAmount(t *testing.T) {
...
+func TestStreamWriteSCTPRollsBackInterleavingMIDsOnSendError(t *testing.T) {
+	testCases := []struct {
+		name      string
+		ppi       PayloadProtocolIdentifier
+		unordered bool
+	}{
+		{
+			name:      "ordered",
+			ppi:       PayloadTypeWebRTCDCEP,
+			unordered: true,
```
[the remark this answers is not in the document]
JoTurk: Good catch, this was a bad copy-paste from the other test.

fippo (approved):
SNAP interaction should be fine and the respective changes look 🍰

ValorZard:
What the heck is going on with the failed test?

```
sctp ERROR: 2026/05/18 01:31:52 [0xc00049ef08] retransmission failure: T1-init
sctp ERROR: 2026/05/18 01:31:52 [0xc00049ec08] retransmission failure: T1-init
sctp ERROR: 2026/05/18 01:31:53 [0xc0001c6f08] retransmission failure: T1-cookie
--- FAIL: TestAssocResetResetsInterleavingCounters (0.21s)
    --- FAIL: TestAssocResetResetsInterleavingCounters/outbound_reset_response_resets_SSN_and_ordered_and_unordered_MIDs (0.21s)
        association_test.go:3292: 
            	Error Trace:	/Users/runner/work/sctp/sctp/association_test.go:3292
            	Error:      	Not equal: 
            	            	expected: 0x0
            	            	actual  : 0x9
            	Test:       	TestAssocResetResetsInterleavingCounters/outbound_reset_response_resets_SSN_and_ordered_and_unordered_MIDs
            	Messages:   	outbound SSN should reset
        association_test.go:3293: 
            	Error Trace:	/Users/runner/work/sctp/sctp/association_test.go:3293
            	Error:      	Not equal: 
            	            	expected: 0x0
            	            	actual  : 0x2
            	Test:       	TestAssocResetResetsInterleavingCounters/outbound_reset_response_resets_SSN_and_ordered_and_unordered_MIDs
            	Messages:   	ordered MID should reset
        association_test.go:3294: 
            	Error Trace:	/Users/runner/work/sctp/sctp/association_test.go:3294
            	Error:      	Not equal: 
            	            	expected: 0x0
            	            	actual  : 0x1
            	Test:       	TestAssocResetResetsInterleavingCounters/outbound_reset_response_resets_SSN_and_ordered_and_unordered_MIDs
            	Messages:   	unordered MID should reset
sctp ERROR: 2026/05/18 01:32:03 Failed to handle chunk: requested non-existent TSN: 2631893422
sctp ERROR: 2026/05/18 01:32:05 [0xc00035a608] client handshake canceled: state=CookieWait
sctp ERROR: 2026/05/18 01:32:05 [0xc0001fd508] client handshake canceled: state=CookieWait
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp ERROR: 2026/05/18 01:32:07 Failed to handle chunk: too many outstanding reconfig requests: 1000
sctp-test ERROR: 2026/05/18 01:32:15 [] released buffer size 1024 should be <= 0
FAIL
FAIL	github.com/pion/sctp	39.905s
?   	github.com/pion/sctp/examples/ping-pong/ping	[no test files]
?   	github.com/pion/sctp/examples/ping-pong/pong	[no test files]
```

JoTurk:
@ValorZard this is just a flaky test, mainly because github runners are slow. not related to this PR.

JoTurk:
@ValorZard I made this test less flaky (with Eventually instead of relying on timing).

FrantaBOT (approved):
LGTM

reassembly_queue.go
```diff
@@ -191,6 +282,48 @@ func (r *reassemblyQueue) push(chunk *chunkPayloadData) bool { //nolint:cyclop
...
+			return true
+		}
+
+		return false
+	}
+
+	if r.orderedMIDMap == nil {
+		r.orderedMIDMap = map[uint32]*chunkSetMID{}
+	}
+	cset = r.orderedMIDMap[chunk.messageIdentifier]
```
sirzooro: 🔴 HIGH: Unbounded Memory Growth via I-DATA / I-FORWARD-TSN with No Stream Limit
Files: reassembly_queue.go, chunk_i_forward_tsn.go

Issue 1 — Unbounded orderedMIDMap and unorderedMIDMap growth

```Go
// reassembly_queue.go – pushIData (ordered path)
cset = r.orderedMIDMap[chunk.messageIdentifier]
if cset == nil {
    cset = newChunkSetMID(chunk.messageIdentifier, chunk.payloadType)
    r.orderedMIDMap[chunk.messageIdentifier] = cset
    r.orderedMID = append(r.orderedMID, cset)
    sortChunkSetsByMID(r.orderedMID)
}
atomic.AddUint64(&r.nBytes, uint64(len(chunk.userData)))
return cset.push(chunk)
```
A remote peer can send an unlimited number of I-DATA chunks with distinct messageIdentifier values (32-bit space = up to ~4 billion distinct keys). Each creates a new map entry and *chunkSetMID in the slice. The only guard is nBytes (payload bytes), but the map entries themselves are unbounded heap objects. An attacker can cause a heap exhaustion DoS by sending billions of 0-byte or tiny-payload I-DATA fragments with unique MIDs. The existing maxReceiveBufferSize only controls nBytes, not the number of in-flight message descriptors.

Recommendation: Enforce a maximum number of concurrent partial MID entries per stream (e.g., cap orderedMIDMap and unorderedMIDMap at a reasonable limit, such as 1024 entries). Abort the association (protocol violation) when exceeded.

Issue 2 — Unbounded orderedMID slice grows with every unseen MID, sorted on every insert

```Go
r.orderedMID = append(r.orderedMID, cset)
sortChunkSetsByMID(r.orderedMID)  // O(n log n) on every push
```
This is also a CPU exhaustion vector: an attacker can hammer ordered I-DATA with sequential MIDs that are never completed, causing O(n log n) sort work on every incoming fragment. For a 32-bit MID space this is quadratic behavior.

Recommendation: Use a sorted-insert instead of append+sort, and cap the slice length.

chunk_i_forward_tsn.go
```diff
@@ -0,0 +1,152 @@
...
+		return err
+	}
+
+	if len(c.raw) < newCumulativeTSNLength {
+		return ErrIForwardTSNChunkTooShort
+	}
+
+	c.newCumulativeTSN = binary.BigEndian.Uint32(c.raw[0:])
+
+	offset := newCumulativeTSNLength
```
sirzooro: 🔴 HIGH: Integer Wrap-around in chunkIForwardTSN.unmarshal Parser Loop
chunk_i_forward_tsn.go, lines 65–77

```Go
offset := newCumulativeTSNLength    // 4
remaining := len(c.raw) - offset
for remaining > 0 {
    s := chunkIForwardTSNStream{}
    if err := s.unmarshal(c.raw[offset:]); err != nil {
        return fmt.Errorf(...)
    }
    c.streams = append(c.streams, s)
    offset += s.length()    // always 8
    remaining -= s.length() // always 8
}
```
remaining and offset are both int. If len(c.raw) is crafted to be very close to maxInt (unlikely in practice but theoretically valid in a 32-bit env), the subtraction remaining -= s.length() can underflow. More practically, if the chunk header's length field is spoofed to indicate a size larger than the actual raw bytes (which are bounds-checked by c.raw = raw[4:length] in the header parser), s.unmarshal(c.raw[offset:]) could receive a too-short slice. The s.unmarshal function does check len(raw) < iForwardTSNEntryLength, so this is partially mitigated — but the error propagation path wraps the error with ErrMarshalStreamFailed and returns, leaving partially parsed c.streams allocated. This is acceptable behavior but should be explicitly noted in comments.

The more concerning form: if s.length() ever changes (it's a method returning a constant today), this becomes exploitable. It's fragile.

Recommendation: Track stream entry count, add a maximum limit (e.g., maxIForwardTSNStreams), and return an error if exceeded.

chunk_i_forward_tsn.go
```diff
@@ -0,0 +1,152 @@
...
+		offset += len(b)
+	}
+
+	c.typ = ctIForwardTSN
+	c.raw = out
+
+	return c.chunkHeader.marshal()
+}
+
+func (c *chunkIForwardTSN) check() (abort bool, err error) {
```
sirzooro, line 125: 🟠 MEDIUM: check() in chunkIForwardTSN Always Returns (true, nil)
chunk_i_forward_tsn.go, line 103

```Go
func (c *chunkIForwardTSN) check() (abort bool, err error) {
    return true, nil
}
```
Compare with chunkForwardTSN.check(), which validates that newCumulativeTSN is not zero and performs basic structural validation. This implementation performs no validation at all. Notably:

A zero newCumulativeTSN in an I-FORWARD-TSN chunk is suspicious and likely invalid.
There's no check that streams entries are well-formed or deduplicated.
A peer can send duplicate stream identifiers (both ordered and unordered flags for the same SI), causing handleForwardTSNForOrderedMID and handleForwardTSNForUnorderedMID to both be called for the same stream, which is likely unintended.
Recommendation: Validate newCumulativeTSN != 0, cap len(c.streams), and check for duplicate (identifier, unordered) pairs.

association.go
```diff
@@ -2007,6 +2191,16 @@ func (a *Association) handleCookieAck() {
 
 // The caller should hold the lock.
 func (a *Association) handleData(chunkPayload *chunkPayloadData) []*packet {
```
sirzooro, line 2210: 🟠 MEDIUM: Protocol Violation Detection Gap — Late useInterleaving Toggle
association.go, handleData

```Go
func (a *Association) handleData(chunkPayload *chunkPayloadData) []*packet {
    if chunkPayload.isIData() != a.useInterleaving {
        if chunkPayload.isIData() {
            a.abortProtocolViolation("received I-DATA without interleaving negotiated")
        } else {
            a.abortProtocolViolation("received DATA with interleaving negotiated")
        }
        return nil
    }
```
This correctly aborts on mismatch. However, useInterleaving is toggled via updateInterleavingState(), which is called from handleInit, handleInitAck, and initWithOutOfBandTokens. In a SNAP (out-of-band) scenario, the state is set synchronously before any data flows. But for a normal INIT/INIT-ACK exchange, there's a window between when a COOKIE-ECHO is received and handleCookieAck completes where useInterleaving may not yet reflect the final negotiated state. If a peer sends I-DATA before ESTABLISHED, the chunk is processed through handleData which checks useInterleaving — and the value may be stale.

Recommendation: Verify that updateInterleavingState() is called before any data chunks can be processed (i.e., ensure it's called on state transition to established).

association_interleaving_options.go
```diff
@@ -0,0 +1,134 @@
...
+package sctp
+
+import "maps"
+
+type interleavingSettings struct {
+	newStreamScheduler InterleavingStreamSchedulerFactory
+	wfqWeights         map[uint16]uint16
+}
+
+func setWeightedFairQueueingStreamScheduler(s *interleavingSettings) {
```
sirzooro, line 13: 🟠 MEDIUM: Scheduler Factory Captured by Closure — Shared State Risk
association_interleaving_options.go, setWeightedFairQueueingStreamScheduler

```Go
func setWeightedFairQueueingStreamScheduler(s *interleavingSettings) {
    s.newStreamScheduler = func() InterleavingStreamScheduler {
        return newWeightedFairQueueingPendingQueuePolicy(s.wfqWeights)
    }
}
```
The closure captures s by reference. s.wfqWeights is the map from the settings struct. If setInterleaving(true) is called, the scheduler is created from the map at that point in time. But if the user later modifies the weights (via a new WithInterleavingWeightedFairQueueingWeight call that clones the settings but the factory still holds s), a race may occur if the factory is called concurrently on a live association.

More critically: cloneInterleavingSettings clones the wfqWeights map but assigns the same newStreamScheduler function reference. That function still holds a reference to the original s, not the clone. So when setInterleaving calls q.newStreamScheduler() on the cloned settings, the factory produces a scheduler using the original settings' weights.

Recommendation: Change the factory to capture a copy of the weights map at the time the factory is set, not at the time it's called. Or re-invoke setWeightedFairQueueingStreamScheduler on the cloned settings during cloneInterleavingSettings.

```Go
// Current (buggy capture):
s.newStreamScheduler = func() InterleavingStreamScheduler {
    return newWeightedFairQueueingPendingQueuePolicy(s.wfqWeights) // `s` is captured
}

// Fix — capture a snapshot:
weights := maps.Clone(s.wfqWeights)
s.newStreamScheduler = func() InterleavingStreamScheduler {
    return newWeightedFairQueueingPendingQueuePolicy(weights)
}
```

association.go
```diff
@@ -1682,17 +1760,99 @@ func getMaxTSNOffset(maxReceiveBufferSize uint32) uint32 {
...
+
+	return nil
+}
+
+func (a *Association) partialReliabilityEnabled() bool {
+	return a.useForwardTSN || a.useIForwardTSN
+}
+
+// The caller should hold the lock.
+func (a *Association) abortProtocolViolation(reason string) {
```
sirzooro, line 1855: 🟡 LOW: additionalInformation in Protocol Violation ABORT Leaks Internal State
association.go, abortProtocolViolation

```Go
func (a *Association) abortProtocolViolation(reason string) {
    a.log.Warnf("[%s] protocol violation: %s", a.name, reason)
    a.willSendAbort = true
    a.willSendAbortCause = &errorCauseProtocolViolation{
        errorCauseHeader:      errorCauseHeader{code: protocolViolation},
        additionalInformation: []byte(reason),
    }
    a.awakeWriteLoop()
}
```
The reason string is sent on-the-wire to the remote peer in the ABORT chunk's Additional Information field. Strings like "received I-DATA without interleaving negotiated" are informational, but the association name (a.name) is logged server-side only, not embedded in the chunk — so this is low risk. However, it's worth ensuring that no future abortProtocolViolation call ever embeds internal addresses, TSN values, or other details into reason that could aid an attacker in fingerprinting the implementation.

Recommendation: Document that reason strings must not contain internal state. Optionally truncate additionalInformation to a fixed max length.

association.go
```diff
@@ -2844,6 +3090,44 @@ func (a *Association) handleForwardTSN(chunkTSN *chunkForwardTSN) []*packet {
...
+	if sna32LTE(chunkTSN.newCumulativeTSN, a.peerLastTSN()) {
+		a.log.Tracef("[%s] sending ack on I-Forward TSN", a.name)
+		a.ackState = ackStateImmediate
+		a.ackTimer.stop()
+		a.awakeWriteLoop()
+
+		return nil
+	}
+
+	for sna32LT(a.peerLastTSN(), chunkTSN.newCumulativeTSN) {
```
sirzooro: 🟡 LOW: handleIForwardTSN TSN Advancement Loop Without Bound
association.go, handleIForwardTSN

```Go
for sna32LT(a.peerLastTSN(), chunkTSN.newCumulativeTSN) {
    a.payloadQueue.pop(true)
}
```
This mirrors the same pattern in handleForwardTSN. If a peer sends an I-FORWARD-TSN with newCumulativeTSN far ahead of peerLastTSN, this loop runs for up to 2^31 iterations (SNA32 wrapping). The existing handleForwardTSN has the same pattern and has presumably been reviewed, but it's worth noting that this unbounded loop could stall the association's write loop goroutine.

Recommendation: Add a maximum iteration count or verify that payloadQueue.pop terminates the loop naturally when the queue is empty, and make that invariant explicit.

reassembly_queue.go
```diff
@@ -369,6 +561,52 @@ func (r *reassemblyQueue) forwardTSNForUnordered(newCumulativeTSN uint32) {
...
+	r.orderedMID = keep
+
+	if sna32LTE(r.nextMID, lastMID) {
+		r.nextMID = lastMID + 1
+	}
+}
+
+func (r *reassemblyQueue) forwardTSNForUnorderedMID(lastMID uint32) {
+	keep := []*chunkSetMID{}
+	for _, set := range r.unorderedMID {
```
sirzooro: 🟡 LOW: forwardTSNForUnorderedMID — Completed Unordered Sets Are Not Dropped if MID <= lastMID
reassembly_queue.go, forwardTSNForUnorderedMID

```Go
for _, set := range r.unorderedMID {
    if sna32LTE(set.mid, lastMID) {
        for _, c := range set.chunks {
            r.subtractNumBytes(len(c.userData))
        }
        continue  // dropped
    }
    keep = append(keep, set)
}
```
This drops completed unordered sets with MID <= lastMID that may not yet have been read by the application. An attacker can force the drop of undelivered data. While the I-FORWARD-TSN mechanism is by design for partial reliability, the implementation should ensure the application has already consumed or explicitly abandoned these, not silently discard data the app was waiting on. This is consistent with RFC 8260 semantics for partial reliability, but the silent data loss may surprise users who have not configured partial reliability for their streams.

sirzooro (commented):
I have asked Copilot with Claude Sonnet 4.6 to perform security review and it found few issues. I am going to do separate manual review after I finish reading RFC document for this feature.

---

Now let me fetch a few specific areas I want to examine more closely for security concerns:Here is a thorough security analysis of this PR as a senior security researcher.

---

## Security Analysis: PR #443 — Stream Schedulers and User Message Interleaving (RFC 8260)

### Summary

This PR implements RFC 8260 (I-DATA / I-FORWARD-TSN chunks, stream schedulers). It is a substantial change (+3422/−176 lines) touching the association state machine, a new pending-queue policy layer, and the reassembly queue. Below are findings organized by severity.

[moved to separate issues]

---

### Summary Table

| Severity | Finding |
|----------|---------|
| 🔴 HIGH | Unbounded `orderedMIDMap`/`unorderedMIDMap` — heap exhaustion DoS |
| 🔴 HIGH | O(n log n) sort on every I-DATA push — CPU exhaustion DoS |
| 🟠 MEDIUM | `chunkIForwardTSN.check()` performs zero validation |
| 🟠 MEDIUM | Closure captures `s` by reference — scheduler factory uses stale/wrong weights |
| 🟠 MEDIUM | `useInterleaving` may be stale during INIT handshake window |
| 🟡 LOW | Protocol violation ABORT leaks internal reason strings on-wire |
| 🟡 LOW | Unbounded TSN advancement loop in `handleIForwardTSN` |
| 🟡 LOW | `forwardTSNForUnorderedMID` silently drops unread completed sets |

The two HIGH-severity findings (unbounded map growth and sort-on-insert) are the most urgent to address before merging, as they are remotely exploitable by any peer who can send I-DATA chunks.

reassembly_queue.go
JoTurk: Interesting finds, I'm going to fix all of them, we do have a lot of similar resources exhaustion issues in pion esp in ICE and RTCP, so i think we just assume the user is able to ban IPs, but it will be better to eventually fix all the similar issues across Pion.

pending_queue.go
```diff
@@ -104,53 +99,359 @@ func (q *pendingQueue) peek() *chunkPayloadData {
...
+
+	popped := streamQueue.pop()
+	if popped != chunkPayload.chunkPayloadData() {
+		return ErrUnexpectedChunkPoppedStream
+	}
+
+	q.virtualTime = math.Max(q.virtualTime, q.chunkFinish[chunkPayload.chunkPayloadData()])
+	delete(q.chunkFinish, chunkPayload.chunkPayloadData())
+	if streamQueue.size() == 0 {
+		delete(q.streamQueues, q.selectedStream)
```
theodorsm, line 365: Shouldn't the `selectedStream` be deleted from `streamFinish` also?

JoTurk: It's harmless, I avoided deleting it to save extra map delete, streamFinish only affects the next push, but Pop already advances virtualTime, so deleting it or keeping it has no effect on the scheduler.

theodorsm (commented):
This is really cool @JoTurk!! 

I have lightly read the RFC and this implementation seems in order:) Interop testing will surface quirks later (if any).

I see there already is some comments here, I could only find one more small thing.

Tag me if you need an approve later.

JoTurk:
@theodorsm THank you so much <3 I'm so excited to just start working on DTLS 1.3 after we release this :)

chunk_i_forward_tsn.go
JoTurk: > A zero newCumulativeTSN in an I-FORWARD-TSN chunk is suspicious and likely invalid.

a zero TSN is valid. I added few checks tho. pushing after few tests.

association.go
JoTurk: No internal state is being leaked :)

JoTurk:
thank you @sirzooro

JoTurk:
I think we'll merge this, if we got new reviews or found new issues, we can just fix them.

PR #447, opened 2025-12-25
fix timers on windows?

JoTurk:
#### Description

apparently we have issues with rack on windows? according to @ValorZard


JoTurk:
obsolete by other graceful shutdown work.

PR #448, opened 2025-12-25
Gracefully shutdown assoc before close

JoTurk:
#### Description
Found this while testing, Now that we set connection deadlines, close might be a bit aggressive.


JoTurk:
I fixed this in different PRs.

Issue #439, opened 2025-11-25
Investigate Congestion Control techniques


philipch07:
With RACK being implemented soon, it would be nice to also investigate congestion control (CC) techniques.

[RFC 9438](https://datatracker.ietf.org/doc/rfc9438/) has been mentioned before and is a classic algorithm but also seems to be one of the oldest stable algorithms for CC. This is something that we definitely do want to add soon, and shouldn't be too difficult to add.

BBRv3 could be worth investigating as well, although it may be a little more difficult to figure out since there isn't an RFC for it.

The crazy thing I want to take a look at sometime is [LeoCC](https://dl.acm.org/doi/pdf/10.1145/3718958.3750491) which totally reinvents things since it's intended for low-earth-orbit (LEO) satellite congestion control which has some different constraints. Thanks to SCP, we should be able to get some really nice benchmarks for SCTP! This will probably be the most difficult/time consuming. It would also probably make our library the first one to implement it for SCTP (at least in the open source world)!

Lanius-collaris:
What do you think of moving the congestion controller out of the methods of `Association` struct? The "burst mitigation" will interfere in pacing too. 
After designing the interface of congestion controller I will "translate" FRCC's kernel module into golang.
> _FRCC: Towards fair and robust congestion control_
> Anup Agarwal, Venkat Arun, and Srinivasan Seshan
> https://www.cs.cmu.edu/~anupa/assets/papers/FRCC-NSDI26.pdf
> https://github.com/108anup/frcc/
### useful features of FRCC
- bounded unfairness
- high utilization (scaling to arbitrarily large bandwidths)
- bounded maximum delay (and delay variation)
- robust to a number of sources of jitter in real networks

FRCC seems to be simpler than BBR v3.
### FRCC's limitation

- convergence time
> Convergence time. Fig. 23 shows 8 flows competing on a 96
> Mbps link with 50 ms RTprop. Each flow enters 90 seconds
> after the previous one and runs for 8 × 90 = 720 seconds.
> Practically, FRCC’s convergence time is slow and takes tens
> of seconds to converge. FRCC takes O(log(BDP)) rounds to
> converge. When there are four flows, a slot is 4 RTTs or 360
> ms (RTT = R + θNG = 50 + 4 ∗ 10), and a round is kNG = 8
> slots or 3 seconds. C.f. a round for BBR is 8∗R = 0.4 seconds.

philipch07:
[...]

PR #458, opened 2026-02-18
allow users to change the congestion control algorithm

Lanius-collaris:
#### Reference issue
Fixes #...
related to https://github.com/pion/sctp/issues/439

### DIFF
--- a/association.go
+++ b/association.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/pion/logging"
 	"github.com/pion/randutil"
+	"github.com/pion/sctp/cc_alg"
 	"github.com/pion/transport/v4/deadline"
 )
 
@@ -313,6 +314,8 @@ type Association struct {
 
 	tlrGoodOps   uint32    // count of TLR ops completed w/o additional loss
 	tlrStartTime time.Time // time of first recovery RTT
+
+	CC1 CongestionController
 }
 
 // Config collects the arguments to createAssociation construction into
@@ -406,6 +409,8 @@ func createAssociation(config Config) *Association {
 		mtu = initialMTU
 	}
 
+	cc1 := cc_alg.CreateReno(config.CwndCAStep)
+
 	tsn := globalMathRandomGenerator.Uint32()
 	assoc := &Association{
 		netConn:              config.NetConn,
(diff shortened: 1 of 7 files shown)


association.go
```diff
@@ -1384,8 +1397,8 @@ func (a *Association) gatherOutbound() ([][]byte, bool) {
 		consumed := false
 
 		rawPackets = a.gatherDataPacketsToRetransmit(rawPackets, &budgetUnits, &consumed)
-		rawPackets = a.gatherOutboundDataAndReconfigPackets(rawPackets, &budgetUnits, &consumed)
 		rawPackets = a.gatherOutboundFastRetransmissionPackets(rawPackets, &budgetUnits, &consumed)
+		rawPackets = a.gatherOutboundDataAndReconfigPackets(rawPackets, &budgetUnits, &consumed)
```
Lanius-collaris: Retransmitting before sending new data can reduce SACKs, should I open one more pull request?

Lanius-collaris:
@philipch07 
If you have time, please try this and give some suggestions. 
These experiments should be useful: 
https://github.com/108anup/frcc/tree/main/experiments

JoTurk (changes requested):
@Lanius-collaris Thank you for working on this, I think configuration CC and more customization is something we should add eventually, after interleaving and after RFC 9260 migration, But i don't think this is mergable in the current shape.

From a library prespactive this does many things:
1. expose CC abstraction.
2. Add FRCC.
3. change ACK/pacing behavior.
4. ALl the other sender-side behavior changes.

And all of these should be split to their own API, and should be standalone and scoped, And after discussion. 

Also I'm worried about turning the library into callbacks/hooks javascriptish nightmare, long term i would rather us moving towards more explicit sans/io style instead of ad hoc hooks.

JoTurk:
@Lanius-collaris I believe the way we should do this is to export more of SCTP internal protocol like dtls and let the advanced users implement SCTP as they wish, adding this interface is dangerous and hard because we can't really enforce the spec with custom CCs algorithms and the interface transition between async and sync and it's really hard to maintain and fix, I'm sorry but I think we'll have to close or reconsider this.

PR #462, opened 2026-03-14
prevent scheduled ACKs from getting cancelled

Lanius-collaris:
#### Description
I hope it works. 😇

#### Reference issue
Fixes https://github.com/pion/sctp/issues/461

### DIFF
--- a/association.go
+++ b/association.go
@@ -179,7 +179,6 @@ func getAssociationStateString(assoc uint32) string {
 type Association struct {
 	bytesReceived uint64
 	bytesSent     uint64
-	unackedBytes  uint32
 
 	lock sync.RWMutex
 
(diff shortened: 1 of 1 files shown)


JoTurk (commented):
This will have extra cost of sending some ACKs slightly earlier in a few bursty cases under high throughput.  But i think this is fine, we can benchmark it later.
We'll need a unit test for this behavior before we can merge it.

Lanius-collaris:
> This will have extra cost of sending some ACKs slightly earlier in a few bursty cases under high throughput.

Two controller in [cc-alg branch](https://github.com/Lanius-collaris/pion-sctp/tree/cc-alg) have `CanSend()` method to tell the caller when to send packets.
https://github.com/Lanius-collaris/pion-sctp/blob/5af9dde48b75c9cc3c6bbb00310976daf4f57e65/cc_alg/frcc.go#L700
https://github.com/Lanius-collaris/pion-sctp/blob/5af9dde48b75c9cc3c6bbb00310976daf4f57e65/cc_alg/reno.go#L71
Do you want to merge it first?

Interleaving nonblocking `Read()` and "blocking" `Read()` ( e.g. `net.Conn.Read` ) can reduce ACKs too.

PR #463, opened 2026-04-04
Expose APIs for handling streamId alloc and re-use

pando-emil:
This commit is a preparation step for making webrtc able to properly determine max number of datachannels that can be used for a specific session.

It also adds an event subscription for when a stream has been completely resetted, so that webrtc knows when it is possible to re-use a streamId.

Corresponding PR in webrtc repo: https://github.com/pion/webrtc/pull/3405


JoTurk:
Thank you @pando-emil sorry I missed this, I'll review both your PRs today.

pando-emil:
@JoTurk did you manage to review this? Just let me know if you have any questions!

JoTurk:
@pando-emil sorry i got busy with some stuff at my work, I'll check this PR. this morning. 
Thank you.

association.go
```diff
@@ -3610,6 +3619,34 @@ func (a *Association) SetMaxMessageSize(maxMsgSize uint32) {
...
+
+// NumOutboundStreams returns the maximum number of outbound streams for this association.
+// The number of outbound streams is determined by looking at the peer's INIT or INIT ACK chunk,
+// so it is not available until the handshake completes.
+func (a *Association) NumOutboundStreams() uint16 {
+	a.lock.RLock()
+	defer a.lock.RUnlock()
+
+	return a.myMaxNumOutboundStreams
+}
```
JoTurk: I think this in-correct for the purpose of this PR because both values are from our direction

JoTurk (changes requested):
Thank you and sorry i took time to get back to you.

association.go
pando-emil: The myMaxNumInboundStreams and myMaxNumOutboundStreams is dynamically changed to the min of our local max and the remotes max in initWithOutOfBandTokens() and handleInitAck(). The added tests in association_test.go should actually verify this.

But maybe we should change name of the variable, "myMax" gets a bit confusing...

JoTurk: > The myMaxNumInboundStreams and myMaxNumOutboundStreams is dynamically changed to the min of our local max and the remotes max in initWithOutOfBandTokens() and handleInitAck(). 

This isn't the problem, the problem is that myMaxNumInboundStreams and myMaxNumOutboundStreams are set to one direction, they are just internally values to hold the remote limits, they are confusing and named incorrectly and we should change them if we were to use them in a public API, And provide an option API to set our own max. 
https://github.com/pion/sctp/blob/2e80e8275db36156b653b55e926d39adef3b14df/association.go#L779-L780

https://github.com/pion/sctp/blob/2e80e8275db36156b653b55e926d39adef3b14df/association.go#L1720-L1721

And how we make init chunks (which is incorrect btw, if we use anything other than MaxUint16 in the future): 

https://github.com/pion/sctp/blob/2e80e8275db36156b653b55e926d39adef3b14df/association.go#L1762-L1763

Another issue is that they are inverted in this PR, NumInboundStreams should have been min(inbound, and the peer's outbound), but currently it's the remote peer's numInboundStreams.

I think if we were to expose APIs or options, we should have, localOS, localMIS, peerOS, peerMIS and drive the max values from min(localOS, peerMIS) and min(localMIS, peerOS).

pando-emil: I wanted to keep the patch to a minimum to not mess with namings etc I don't know the original intentions behind. But I do agree with your comments. I'll try to rework the internal states and update the PR.

But I think the addition of an API to set the local max can be saved for a separate PR, that is a new feature.

JoTurk: Sounds good, thank you.

JoTurk:
@pando-emil  I'm thinking about this again, because in interleaving i want to expose an API so the end user can tell if interleaving is enabled and negotiated or not, and since we're going to export multiple methods that are only meaningful after handshake I think we should have a metadata struct and a way to register a callback or even blocking function that returns all the assoc metadata.

pando-emil:
@JoTurk that sounds reasonable! I'll adopt to those changes as soon as they are merged. In the meantime I have updated with a reworked representation within association.

JoTurk:
@pando-emil the metadata API is merged and released btw if you want to update your PR to adds stats to it.

pando-emil:
@JoTurk Well, I guess it is hard to claim I did the adoption ASAP as I promised. But now, finallly! :)

pando-emil:
@JoTurk I saw your comment in #485, I didn't realize I messed up the test until now, I apologize for that! Should be fixed now.

pando-emil:
And I agree with @idy, whatever merge method of the 2 PRs you prefer works for me!

PR #466, opened 2026-05-17
disable interleaving until pion 4.3

JoTurk:
#### Description
based on https://github.com/pion/sctp/pull/443 
Disables interleaving until we release 4.3 and enable it by default.


JoTurk:
Let's do it without a flag :)

PR #469, opened 2026-06-20
Fix endless loop in processSelectiveAck

sirzooro:


association.go
```diff
@@ -2583,8 +2583,8 @@ func (a *Association) processSelectiveAck(selectiveAckChunk *chunkSelectiveAck)
 
 	// Mark selectively acknowledged chunks as "acked"
 	for _, g := range selectiveAckChunk.gapAckBlocks {
-		for i := g.start; i <= g.end; i++ {
-			tsn := selectiveAckChunk.cumulativeTSNAck + uint32(i)
+		for i := uint32(g.start); i <= uint32(g.end); i++ {
+			tsn := selectiveAckChunk.cumulativeTSNAck + i
```
JoTurk, line 2587: nice find can we add a regression test for {start: 65535, end: 65535} if possible?

JoTurk (approved):

association.go
sirzooro: Added

PR #471, opened 2026-06-24
Set cap of ReassemblyQueue based on rwnd size

cnderrauber:
#### Description

A lost reliable message or slow application reader could cause
receiver exceeds the fixed 1024 cap of reassemble queue when
network capacity is high. Adjusting this limit to a value based
on the rwnd size to mitigate this issue.

### DIFF
--- a/association_test.go
+++ b/association_test.go
@@ -1122,12 +1122,14 @@ func TestAssociationIForwardTSNDuplicateEntriesUseLargestMID(t *testing.T) {
 }
 
 func TestAssociationInterleavingProtocolViolationMIDLimit(t *testing.T) {
-	assoc := createTestAssociation(t, Config{})
+	const limit = maxReassemblyQueueMIDEntries
+	// receive buffer sized so the derived MID cap is exactly limit
+	assoc := createTestAssociation(t, Config{MaxReceiveBufferSize: reassemblyMIDLimitBytesPerEntry * limit})
 	assoc.useInterleaving = true
 	assoc.payloadQueue.init(0)
 	assoc.setState(established)
 
-	for i := range maxReassemblyQueueMIDEntries {
+	for i := range limit {
 		packets := assoc.handleData(&chunkPayloadData{
 			iData:                  true,
 			messageIdentifier:      uint32(i), //nolint:gosec // bounded by maxReassemblyQueueMIDEntries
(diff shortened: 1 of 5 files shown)


JoTurk:
@cnderrauber this hard cap was added because of this security review https://github.com/pion/sctp/pull/443#discussion_r3275515150 by @sirzooro  we can't just switch to a_rwnd because header-only I-DATA consumes 0 a_rwnd so this exposes us to DoS attacks 

the RFC doesn't suggest any specific limit here but i saw the comment is reasonable enough to add a limit

>   It should be noted that the application has to consent that it is
   willing to do the more complex reassembly support required for user
   message interleaving.  When doing so, an application has to provide a
   reassembly buffer for each incoming stream.  It has to protect itself
   against these buffers taking too many resources.  If user message
   interleaving is not used, only a single reassembly buffer needs to be
   provided for each association.  But the application has to protect
   itself for excessive resource usages there too.
 >  RFC 8260 
   
 Maybe we can just change the limit to a much bigger number?

sirzooro:
Some limit should be in place, to prevent unbounded memory growth mentioned in linked issue. We can increase this default limit and make it configurable. Additionally we could calculate size of all fragments waiting in reassembly queue and abort when it exceeds some limit (also configurable). Big number of allowed waiting fragments combined with reasonable size limit should help with case when there are many small fragments waiting.

boks1971:
Configurable limits (defaults higher than current) + size check sounds like a good approach to me.

boks1971 (commented):
lgtm! Would be good to more 👀 on this.

association_test.go
```diff
@@ -1122,12 +1122,14 @@ func TestAssociationIForwardTSNDuplicateEntriesUseLargestMID(t *testing.T) {
 }
 
 func TestAssociationInterleavingProtocolViolationMIDLimit(t *testing.T) {
-	assoc := createTestAssociation(t, Config{})
+	const limit = maxReassemblyQueueMIDEntries
+	// receive buffer sized so the derived MID cap is exactly limit
+	assoc := createTestAssociation(t, Config{MaxReceiveBufferSize: reassemblyMIDLimitBytesPerEntry * limit})
```
JoTurk, line 1127: note that an attacker can open up to 16 streams before accept is called, so the actual limit per an assoc for an attacker here is changed to `262,144` but at this point it's reasonable to add memory-based limit later.

JoTurk (approved):
thank you.

PR #472, opened 2026-06-26
Configurable optin reassembly queue limit for DATA / I-DATA and make it optional / optin

JoTurk:
#### Description
After discussion in https://discord.com/channels/1352636971591274548/1359189787768258671/1519753476983357531 we decided to make any arbitrary limits or limits not mandated by RFC optional even if they lead to DoS.
This makes the I-DATA queue limit optin (none by default) and applies the same optional limit to DATA queues.
At pion/webrtc we should export a single API to set all the similar limits across Pion at once, and maybe provide profiles, or way to implement fair.

related to https://github.com/pion/sctp/pull/471

### DIFF
--- a/association.go
+++ b/association.go
@@ -270,17 +270,18 @@ type Association struct {
 	recvZeroChecksum        bool
 
 	// Congestion control parameters
-	maxReceiveBufferSize uint32
-	maxMessageSize       uint32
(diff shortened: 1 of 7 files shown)


boks1971 (approved):

cnderrauber (approved):
Looks good! Thank you!!

PR #473, opened 2026-06-27
Export ClientContext

HashimTheArab:
Export `ClientContext(ctx, opts...)` so callers can bound or cancel SCTP connection setup. This enables a follow-up `SCTPTransport.StartContext` API in pion/webrtc without waiting for the full SCTP handshake when a connection attempt is canceled.

The context controls establishment only. Canceling it after successful setup leaves the association open. An already-canceled context starts no association and leaves the supplied connection open. Cancellation during setup returns `ctx.Err()` and closes the association in the background, including the SNAP path, so a blocking connection shutdown cannot stall the caller.

Use the maintainer-requested `ClientContext` name so the API remains consistent when the transitional `WithOptions` constructors are renamed in the next major version. Existing constructors continue to use a background context.

Validation: full Go race suite, focused tests for cancellation/deadlines and connection cleanup, and CI-pinned golangci-lint v2.10.1.

### DIFF
--- a/association.go
+++ b/association.go
@@ -437,11 +437,11 @@ func Client(config Config) (*Association, error) {
 
 // ClientWithOptions opens a SCTP stream over a conn.
 func ClientWithOptions(opts ...ClientOption) (*Association, error) {
-	return createClientWithOptionsWithContext(context.Background(), opts...)
+	return ClientWithOptionsContext(context.Background(), opts...)
 }
 
 func createClientWithContext(ctx context.Context, config Config) (*Association, error) {
-	return createClientWithOptionsWithContext(ctx, config)
+	return ClientWithOptionsContext(ctx, config)
 }
 
 func createSNAPAssociation(config *Config) (*Association, error) {
@@ -464,7 +464,10 @@ func createSNAPAssociation(config *Config) (*Association, error) {
 	return assoc, nil
 }
 
-func createClientWithOptionsWithContext(ctx context.Context, opts ...ClientOption) (*Association, error) {
+// ClientWithOptionsContext opens a SCTP stream over a conn.
+// If ctx is canceled before the SCTP handshake completes, the association is
+// closed and ctx.Err() is returned.
+func ClientWithOptionsContext(ctx context.Context, opts ...ClientOption) (*Association, error) {
 	config, err := buildClientConfig(opts...)
 	if err != nil {
 		return nil, err
--- a/association.go
+++ b/association.go
@@ -437,11 +437,11 @@ func Client(config Config) (*Association, error) {
 
 // ClientWithOptions opens a SCTP stream over a conn.
 func ClientWithOptions(opts ...ClientOption) (*Association, error) {
-	return ClientWithOptionsContext(context.Background(), opts...)
+	return ClientContext(context.Background(), opts...)
 }
 
 func createClientWithContext(ctx context.Context, config Config) (*Association, error) {
-	return ClientWithOptionsContext(ctx, config)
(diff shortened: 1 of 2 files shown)


HashimTheArab:
@JoTurk

JoTurk:
Thank you, I'll try to review it in the morning, one thing tho, we shouldn't add withOptions to the name as this is a temporary name until we release the next major version and move to functional option only.
Also can you explain a use case for this API.

HashimTheArab:
Got it, thanks, and the reason is what I explained in the PR description, a follow-up PR will need this to add SCTPTransport.StartContext which is for the same reason of https://github.com/pion/webrtc/pull/3371

HashimTheArab:
@JoTurk let me know if anything else is needed, thanks!

JoTurk:
Sounds good, I'll try to check this tonight

HashimTheArab:
bump

JoTurk:
Sorry i lost track of this, and i remembered it while i was reviewing https://github.com/pion/webrtc/pull/3489

HashimTheArab:
hey sorry to bother, we really need this merged if possible

JoTurk (changes requested):
Sorry i forgot to comment, createClientWithOptionsWithContext isn't safe as a public API. some reasons:

1. context cancellation can block forever on assoc.Close call, please look at blockingCloseConn for more context.
2. if snap is enabled we skip the context createSNAPAssociation.

once we fix these issues i think we can expose it as a public API.

HashimTheArab:
should be fixed

client_context_test.go
```diff
@@ -0,0 +1,182 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
```
JoTurk: Can you please move this test to association_test, we try to avoid adding new test files in '/' unless they test an existing file name.

association.go
```diff
@@ -1058,6 +1068,18 @@ func (a *Association) Close() error {
...
+func (a *Association) closeAsync() {
+	go func() {
+		_ = a.netConn.SetReadDeadline(time.Now())
+
+		if err := a.Close(); err != nil {
+			a.log.Warnf("[%s] failed to close association: %v", a.name, err)
+		}
+	}()
+}
+
```
JoTurk, line 1084: This can cause a gorotune leak if close blocks (can happen with ice-tcp and also with dtls shutdown notification), but shouldn't try to attempt to fix this here, it should be fixed in the new DTLS write path. but we should add a WriteDeadline here at least.

JoTurk (commented):
Ty just two last comments.

HashimTheArab:
resolved, thanks!

JoTurk (approved):

JoTurk:
Thank you, I'll merge after the CI.

glancist:
Thanks mommy

JoTurk:
@glancist That's a pretty inappropriate thing to say, I assume English isn't your first language

glancist:
@JoTurk My apologies for the tone, that came out wrong. I just wanted to say I'm grateful this got merged, thanks for taking the time to review it!

NopeNotDark:
Mom can you lock this PR please, there's a weirdo by the name of glancist here

JoTurk:
@NopeNotDark banned for 30 days and reported, @HashimTheArab Can you please ask your friends to stop

HashimTheArab:
Im very sorry about that I mentioned the PR in a private group and did not think they would leave those comments

PR #474, opened 2026-07-02
Update readme

Ibn-Butt-Toota:
#### Description
I have updated the readme to include RFC 8260, as well as adding RACK and Adaptive Burst Mitigation from the table to a separate section so it's more clear at a glance.

#### Reference issue
n/a


sirzooro:
"Potential future implementations" table below contains few lines marked as implemented, probably they should be removed? Just make sure to include relevant links to specifications in other parts of this document.

Also add link to SNAP https://datatracker.ietf.org/doc/draft-hancke-tsvwg-snap/

Ibn-Butt-Toota:
> "Potential future implementations" table below contains few lines marked as implemented, probably they should be removed? Just make sure to include relevant links to specifications in other parts of this document.
> 
> Also add link to SNAP https://datatracker.ietf.org/doc/draft-hancke-tsvwg-snap/

Hello, good ideas. I set the SNAP link to a snapshot of the draft that the current implementation is based on.

JoTurk (approved):

Issue #12, opened 2019-01-23
Add simple PMTU calculation with heartbeats


kc5nra:
Calculate the maximum MTU size that can travel the entire path machine1<->machine2.

The easiest way is probably to craft heartbeat messages starting at the maximum value until they are successfully returned.  Potentially listen to ICMP PacketTooBig

Sean-Der:
@enobufs is this still an issue/something that needs to be done?

enobufs:
SCTP RFC actually mentions RFC4821 which recommends using the heartbeat. It is an interesting approach.  The current implementation does not fully support heartbeat. In WebRTC context, we have heartbeat at ICE layer, and I am not sure how critical it is not having heartbeat at SCTP layer..

I think we need to investigate how other WebRTC libraries deal with path MTU discovery.

nils-ohlmeier:
> I think we need to investigate how other WebRTC libraries deal with path MTU discovery.

AFAIK libwebrtc actually has the PMTU hard coded to a safe low value (I would have to lookup the exact value).

And yes the two options are either on the SCTP or the ICE level. 

The ICE PMTU would have the advantage that the same information could get passed to the RTP stack as well.
But the problem is that for ICE there was only a draft for figuring out the PMTU, never an RFC. And AFAIK there is no other implementation which supports the ICE way.
The SCTP approach is way more interoperable.

PR #476, opened 2026-07-14
Set default MTU size to 1191

JoTurk:
#### Description
some Raspberry PI connect users reported that pion/sctp can't send video over SCTP while Tailscale is enabled (Tailscale drops packets larger than 1280 bytes), with DTLS record header, cipher nonce, cipher auth tag, and IP header we can easily go over 1280 with 1228 as the default.

This sets the default MTU size to safer value[1] of 1191 instead of 1228 and account for chunk padding.

1. https://issues.webrtc.org/issues/42222640

I'll try to get PMTU implemented soon #12

### DIFF
--- a/association.go
+++ b/association.go
@@ -70,13 +70,43 @@ var (
 
 const (
 	receiveMTU            uint32 = 8192 // MTU for inbound packet (from DTLS)
-	initialMTU            uint32 = 1228 // initial MTU for outgoing packets (to DTLS)
+	initialMTU            uint32 = 1191 // initial MTU for outgoing packets (to DTLS)
 	initialRecvBufSize    uint32 = 1024 * 1024
-	commonHeaderSize      uint32 = 12
-	dataChunkHeaderSize   uint32 = 16
-	iDataChunkHeaderSize  uint32 = 20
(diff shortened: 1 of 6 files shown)


fippo (approved):
ah you read https://p2claw.com/blog/2026-06-09-the-ipad-was-on-tailscale/ ? :-)

JoTurk:
@fippo yes, when you linked it in the Discord :)  @Sean-Der mentioned that MTU could be the issue, and it made sense to check pion's value now https://discord.com/channels/1352636971591274548/1524781900395450591/1526273095084806248 it's funny because webrtc-rs inherited 1228 from pion (tbh i didn't remember correctly and I thought Pion's default was actually 1200).

we even went over 1228 thanks to the fact that we didn't account for chunk padding lol.

JoTurk:
(tagged this in a new minor release instead of a patch release, because it's kinda of a behavior change)

JoTurk:
@fippo nvm i remembered correctly our mtu was actually 1200 we just overwrite in webrtc https://github.com/pion/webrtc/pull/3476

PR #480, opened 2026-07-26
Prevent partial sack processing

JoTurk:
#### Description
Validate that full range exists in the inflight queue to prevent partial pops
left over from an invalid SACK that causes the cumulativeTSNAckPoint to be updated 
incorrectly and causes weird state and race conditions.
#### Reference issue
I thought this was caused by a broken test when I saw it first in the CI but it happens a few times.

When a sack contained a gap referencing a missing tsn, we could:
1. Remove cumulatively acknowledged packets from the in-flight queue.
2. Or, encounter the missing TSN and return an error.
3, Or, fail to update the cumulative ACK point.

shutdown then waited forever for data to drain, causing the five-minute timeout and goroutine leaks.

> sctp ERROR: 2026/07/26 03:37:30 [0xc000208388] retransmission failure: T1-cookie
sctp ERROR: 2026/07/26 03:37:39 Failed to handle chunk: requested non-existent TSN: 3597224921
--- FAIL: TestAssociation_ShutdownDuringWrite (301.06s)
    association_test.go:4892: 
        	Error Trace:	/Users/runner/work/sctp/sctp/association_test.go:4892
        	Error:      	Received unexpected error:
        	            	context deadline exceeded
        	Test:       	TestAssociation_ShutdownDuringWrite
        	Messages:   	timed out waiting for a1 shutdown to complete
    association_test.go:3470: 
        	Error Trace:	/Users/runner/work/sctp/sctp/association_test.go:3470
        	            				/Users/runner/hostedtoolcache/go/1.24.13/arm64/src/testing/testing.go:1211
        	            				/Users/runner/hostedtoolcache/go/1.24.13/arm64/src/testing/testing.go:1445
        	            				/Users/runner/hostedtoolcache/go/1.24.13/arm64/src/testing/testing.go:1786
        	            				/Users/runner/hostedtoolcache/go/1.24.13/arm64/src/runtime/panic.go:636
        	            				/Users/runner/hostedtoolcache/go/1.24.13/arm64/src/testing/testing.go:1041
        	            				/Users/runner/work/sctp/sctp/association_test.go:4892
        	Error:      	goroutine leak
        	Test:       	TestAssociation_ShutdownDuringWrite
        	Messages:   	leaked: 6

PR #481, opened 2026-07-26
Ack data in SHUDOWN-SENT

JoTurk:
#### Description
A gap i found while fixing https://github.com/pion/sctp/pull/480 
accept and immediately acknowledge data while in SHUTDOWN-SENT, then send a SHUTDOWN response, and restart T2 to prevent shutdown deadlocks.

### DIFF
--- a/association.go
+++ b/association.go
@@ -1599,7 +1599,7 @@ func (a *Association) gatherOutboundShutdownPackets(rawPackets [][]byte) ([][]by
 		a.willSendShutdown = false
 
 		shutdown := &chunkShutdown{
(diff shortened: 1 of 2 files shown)


boks1971 (approved):

Ibn-Butt-Toota (approved):

JoTurk:
I found few other issues and experimented with fixes here but i reset the commit to the original and going to open two PRs to fix the other issues.

JoTurk:
the issues were found while i was doing manual interop testings against dcsctp.

PR #482, opened 2026-07-27
Process shutdown cumulative ACKs

JoTurk:
#### Description
One of the issues I found while i was interop testing #481 with other libraries this refactors SACK ack handling so a SHUTDOWN's cumulative TSN ACK can release acknowledged DATA and update state. and enters shutdownReceived state. This improves graceful shutdown for the other peer side.

PR #483, opened 2026-07-27
Handle graceful shutdown state transitions

JoTurk:
#### Description
this commit was first added to #481 while i was interop testing it, and continues from #482 
This is the final fix for the graceful shutdown and makes the state linear ESTABLISHED => SHUTDOWN-PENDING => drain queued/inflight data => SHUTDOWN-SENT => SHUTDOWN-COMPLETE => CLOSED.
it also handles some edge cases while doing graceful shutdown, like if both peers shutdown simultaneously. and prevents late DATA from restarting shutdown.

The diff seems noisy because i added many tests.

### DIFF
--- a/association.go
+++ b/association.go
@@ -265,6 +265,7 @@ type Association struct {
 	willSendShutdown         bool
 	willSendShutdownAck      bool
 	willSendShutdownComplete bool
+	shutdownCompletePending  bool
 
 	willSendAbort      bool
 	willSendAbortCause errorCause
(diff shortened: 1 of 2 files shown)


Ibn-Butt-Toota (approved):
This is awesome but it makes me scared of the association refactor down the line. The file is comically large right now. Also nice work with the tests as well!

JoTurk:
I'll try to merge all the pending prs then i'll start refactoring it and moving stuff to internal :)

PR #485, opened 2026-08-07
Make SCTP stream reset and reuse generation-safe

idy:
## What changed

- expose reset-completion notifications and negotiated stream limits
- retain stream ownership until both reset directions have completed
- keep delayed acknowledgements and retransmitted reset requests from affecting a reused stream ID
- preserve established associations when a duplicate INIT arrives
- separate handshake retransmission limits from data retransmission limits
- copy configured inbound and outbound stream limits into client and server association options

## Why

WebRTC needs to reuse SCTP stream IDs after a DataChannel closes, but an ID is not safe to reuse until the reset procedure has completed. Delayed acknowledgements or a retransmitted reset request from the previous stream generation must also not close or mutate the new generation.

The handshake retry limit is separated because using the data retransmission limit for INIT/COOKIE retries can either terminate association setup too early or make data retransmission behavior unexpectedly permissive.

## Relationship to #463

This is an alternative implementation that covers the stream-limit and reset-completion goals of #463 and adds the lifecycle and retransmission handling needed by the corresponding WebRTC integration.

This PR is not intended to invalidate the work in #463. If maintainers prefer to continue with #463, please feel free to close this PR and reuse or adapt any tests and fixes that are useful.

## Validation

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `staticcheck .`

### DIFF
--- a/association.go
+++ b/association.go
@@ -283,6 +283,7 @@ type Association struct {
 	reconfigs                    map[uint32]*chunkReconfig
 	reconfigRequests             map[uint32]*paramOutgoingResetRequest
 	onStreamResetCompleteHandler func(streamID uint16)
+	streamResetStates            map[uint16]streamResetState
 
 	// Non-RFC internal data
 	sourcePort              uint16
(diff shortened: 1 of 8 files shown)


JoTurk:
hello, can this effort merged with @pando-emil in #463 as single PR or after their PR is merged (currently the CI is failing and requires a review)?

idy:
Yes, absolutely. We are happy to follow whichever approach the maintainers prefer. We can combine this work with @pando-emil's #463 into a single PR, or rebase the additional fixes and tests on top after #463 is merged. Please let us know which path is easiest to review, and we will adapt #485 accordingly.

idy:
Superseded by #494, which is opened from a branch scoped to the stream reset/reuse changes only. The duplicate-INIT and handshake retransmission limit fixes that were bundled here now live on separate branches.

PR #487, opened 2026-08-09
Expose User Initiated Abort through errors.Is

### LINKED ISSUE #486: Expose typed ABORT causes for errors.Is/errors.As
### Environment

- `github.com/pion/sctp` v1.11.1
- `github.com/pion/webrtc/v4` v4.2.18
- Also checked current main branches:
  - sctp `94ddb36b4e89`
  - webrtc `381746d5232c`

### What did you do?

In a standalone Pion-to-Pion loopback:

1. Create two `webrtc.PeerConnection` instances.
2. Open a DataChannel and wait until it is open on both sides.
3. Register `DataChannel.OnError` on the receiving side.
4. Call `PeerConnection.Close()` on the other side.
5. Inspect the error delivered to `OnError`.

The reproducer uses only public Pion APIs.

<details>
<summary>Standalone reproducer</summary>

```go
package abortrepro

import (
	"errors"
	"testing"
	"time"

	"github.com/pion/sctp"
	"github.com/pion/webrtc/v4"
)

func TestRemotePeerConnectionCloseReportsOnlyErrChunk(t *testing.T) {
	left, err := webrtc.NewPeerConnection(webrtc.Configuration{})
	if err != nil {
		t.Fatal(err)
	}
	right, err := webrtc.NewPeerConnection(webrtc.Configuration{})
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() {
		_ = left.Close()
		_ = right.Close()
	})

	remoteOpen := make(chan struct{})
	remoteError := make(chan error, 1)
	right.OnDataChannel(func(dc *webrtc.DataChannel) {
		dc.OnOpen(func() { close(remoteOpen) })
		dc.OnError(func(err error) { remoteError <- err })
	})
	local, err := left.CreateDataChannel("repro", nil)
	if err != nil {
		t.Fatal(err)
	}
	localOpen := make(chan struct{})
	local.OnOpen(func() { close(localOpen) })

	offer, err := left.CreateOffer(nil)
	if err != nil {
		t.Fatal(err)
	}
	leftGathered := webrtc.GatheringCompletePromise(left)
	if err := left.SetLocalDescription(offer); err != nil {
		t.Fatal(err)
	}
	<-leftGathered
	if err := right.SetRemoteDescription(*left.LocalDescription()); err != nil {
		t.Fatal(err)
	}
	answer, err := right.CreateAnswer(nil)
	if err != nil {
		t.Fatal(err)
	}
	rightGathered := webrtc.GatheringCompletePromise(right)
	if err := right.SetLocalDescription(answer); err != nil {
		t.Fatal(err)
	}
	<-rightGathered
	if err := left.SetRemoteDescription(*right.LocalDescription()); err != nil {
		t.Fatal(err)
	}

	select {
	case <-localOpen:
	case <-time.After(5 * time.Second):
		t.Fatal("local DataChannel did not open")
	}
	select {
	case <-remoteOpen:
	case <-time.After(5 * time.Second):
		t.Fatal("remote DataChannel did not open")
	}

	if err := left.Close(); err != nil {
		t.Fatal(err)
	}
	select {
	case err := <-remoteError:
		if !errors.Is(err, sctp.ErrChunk) {
			t.Fatalf("OnError error = %v; want sctp.ErrChunk", err)
		}
		t.Logf("OnError error: %v", err)
	case <-time.After(5 * time.Second):
		t.Fatal("remote DataChannel OnError was not called")
	}
}
```

</details>

### What happened?

The remote DataChannel receives an error similar to:

```text
abort chunk, with following errors: (User Initiated Abort: )
```

The error matches `sctp.ErrChunk`:

```go
errors.Is(err, sctp.ErrChunk) == true
```

However, the User Initiated Abort cause (RFC 9260 cause code 12) is represented by private types and is flattened into formatted text. Callers cannot distinguish it from an ABORT caused by a protocol violation using `errors.Is` or `errors.As`.

Parsing the error string would be brittle. Treating every `sctp.ErrChunk` as an expected close could hide real protocol failures.

### What did you expect?

A stable typed signal for the received ABORT cause, for example either:

```go
errors.Is(err, sctp.ErrUserInitiatedAbort)
```

or:

```go
var abortErr *sctp.AbortError
if errors.As(err, &abortErr) {
	// Inspect typed cause codes.
}
```

The exact API shape is not important. The requirement is that the received cause code survives wrapping and can be inspected without parsing text. Callers can then decide their own close policy.

### Suggested direction

Preserve ABORT cause information in the error returned by `Association.handleAbort`, while continuing to wrap `ErrChunk` for compatibility.

A sentinel for User Initiated Abort would solve the immediate case. A structured abort error containing typed cause codes may be more extensible.

### Related issues

- #251 exposed the general SCTP error values, but `ErrChunk` still covers every ABORT cause.
- #425 updates RFC 9260 error-cause parsing, but the concrete cause types remain private and are not exposed through the returned error chain.

kaptinlin:
## Summary

- add `ErrUserInitiatedAbort` for received ABORT chunks containing RFC 9260 cause code 12
- preserve `ErrChunk` compatibility while exposing the specific cause through standard error wrapping
- keep other ABORT causes distinct without exporting a broader structured error API

## Testing

- `GOTOOLCHAIN=go1.24.0 GOMAXPROCS=4 go test -race ./...`
- `GOTOOLCHAIN=go1.25.0 GOMAXPROCS=4 go test -race . -run "^(TestAssocAbort|TestAssociation_Abort)$" -count=10`
- `GOTOOLCHAIN=go1.25.0 GOMAXPROCS=4 go vet ./...`
- `golangci-lint v2.10.1 run --allow-parallel-runners`

Fixes #486

### DIFF
--- a/association.go
+++ b/association.go
@@ -68,23 +68,6 @@ var (
 	ErrTooManyReconfigRequests    = errors.New("too many outstanding reconfig requests")
 )
 
-type abortError struct {
-	message               string
-	hasUserInitiatedAbort bool
-}
-
-func (e *abortError) Error() string {
-	return e.message
-}
-
-func (e *abortError) Is(target error) bool {
-	return target == ErrUserInitiatedAbort && e.hasUserInitiatedAbort
-}
-
-func (e *abortError) Unwrap() error {
-	return ErrChunk
-}
-
 const (
 	receiveMTU            uint32 = 8192 // MTU for inbound packet (from DTLS)
 	initialMTU            uint32 = 1191 // initial MTU for outgoing packets (to DTLS)
@@ -3371,10 +3354,11 @@ func (a *Association) handleAbort(c *chunkAbort) error {
 
 	_ = a.close()
 
-	return &abortError{
-		message:               fmt.Sprintf("[%s] %s: %s", a.name, ErrChunk, errStr.String()),
-		hasUserInitiatedAbort: hasUserInitiatedAbort,
+	if hasUserInitiatedAbort {
+		return fmt.Errorf("[%s] %w: %w: %s", a.name, ErrChunk, ErrUserInitiatedAbort, errStr.String())
 	}
+
+	return fmt.Errorf("[%s] %w: %s", a.name, ErrChunk, errStr.String())
 }
 
 // createForwardTSN generates ForwardTSN chunk.
--- a/association.go
+++ b/association.go
@@ -68,6 +68,23 @@ var (
 	ErrTooManyReconfigRequests    = errors.New("too many outstanding reconfig requests")
 )
 
+type abortError struct {
+	message               string
+	hasUserInitiatedAbort bool
+}
+
+func (e *abortError) Error() string {
+	return e.message
+}
+
+func (e *abortError) Is(target error) bool {
+	return target == ErrUserInitiatedAbort && e.hasUserInitiatedAbort
+}
+
+func (e *abortError) Unwrap() error {
+	return ErrChunk
+}
+
 const (
 	receiveMTU            uint32 = 8192 // MTU for inbound packet (from DTLS)
 	initialMTU            uint32 = 1191 // initial MTU for outgoing packets (to DTLS)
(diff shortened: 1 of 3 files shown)


association.go
```diff
@@ -68,6 +68,23 @@ var (
...
+	return e.message
+}
+
+func (e *abortError) Is(target error) bool {
+	return target == ErrUserInitiatedAbort && e.hasUserInitiatedAbort
+}
+
+func (e *abortError) Unwrap() error {
+	return ErrChunk
+}
```
JoTurk: Can we use standard errors wrapping?  thank you.

kaptinlin: Yes, updated in 26e9a15 to use fmt.Errorf with multiple %w verbs and removed the custom abortError. Both ErrChunk and ErrUserInitiatedAbort remain matchable through errors.Is, while other ABORT causes still only match ErrChunk. Thank you.

JoTurk (approved):
Thank you, sorry i lost track of this. Thank you for removing the custom error wrapper after my comment, it does break direct `errors.Unwrap` use. but calling `errors.Unwrap` is anti-pattern, and we shouldn't add code to support this use.

PR #491, opened 2026-09-01
Omit unordered streams from FORWARD TSN

maum-meliora:
## Problem

`createForwardTSN` reports every skipped TSN in the advanced range,
including unordered DATA. Unordered DATA does not consume stream
sequence numbers, so its `streamSequenceNumber` is not a meaningful SSN
— copying it into the chunk can tell the receiver to skip an ordered
message that was never sent.

RFC 3758 Section 3.2 forbids this, and the requirement is already quoted
in this package, on `chunkForwardTSNStream.sequence`:

> This field MUST NOT report TSN's corresponding to DATA chunks that are
> marked as unordered.

## Change

Skip unordered chunks when building the stream list.

`createIForwardTSN` is intentionally left unchanged: RFC 8260 gives
I-FORWARD TSN entries an ordered flag and a message identifier rather
than an SSN, so unordered messages belong there.

## Testing

- `createForwardTSN` omits unordered streams, including when unordered
  and ordered DATA share a stream identifier; the first case is asserted
  on the marshaled bytes.
- A subtest pins the deliberate asymmetry: `createIForwardTSN` keeps
  both ordered and unordered entries.
- `go test ./...`, `go test -race ./...`, golangci-lint v2.10.1


JoTurk (approved):
thank you

PR #492, opened 2026-09-01
Handle pending scheduler pop failures

maum-meliora:
## Problem

`popPendingDataChunksToSend` treats `pendingQueue.pop` as infallible,
but the pop is delegated to the stream scheduler policy and can return
an error. When it does, the chunk stays in the pending queue while the
association has already acted as if it left:

- a DATA chunk is assigned a TSN and pushed onto the inflight queue while
  it is still in the pending queue, and its length is subtracted from
  rwnd;
- a reset marker's stream identifier is queued for a reset that never
  had its queue entry removed;
- the burst budget consumed for that chunk is not given back, and the
  same entry is peeked again on the next pass;
- with nothing sent, the zero window probe branch can then pick the same
  entry.

## Change

Return whether the pop succeeded and stop the pass when it fails,
before assigning TSNs, consuming rwnd, or queueing a reset. The burst
budget consumed for the failed chunk is restored, chunks moved
successfully earlier in the pass are kept, and the zero window probe is
suppressed for that pass.

The default weighted fair queueing scheduler does not fail this pop, so
this only affects associations using a custom scheduler.

## Testing

- A scheduler stub that rejects the pop covers the DATA, zero window
  probe, and reset marker paths: no TSN is assigned, no chunk is sent,
  no reset is queued, and the burst budget is restored.
- A later-failure case: the first chunk is kept and only the failed
  chunk's budget is restored.
- `go test ./...`, `go test -race ./...`, golangci-lint v2.10.1


association.go
```diff
@@ -3791,11 +3796,14 @@ func (a *Association) popPendingDataChunksToSend( //nolint:cyclop,gocognit
 
 			dataLen := uint32(len(chunkPayload.userData)) //nolint:gosec // G115
 			if dataLen == 0 {
-				sisToReset = append(sisToReset, chunkPayload.streamIdentifier)
 				err := a.pendingQueue.pop(chunkPayload)
 				if err != nil {
 					a.log.Errorf("failed to pop from pending queue: %s", err.Error())
+					pendingQueuePopFailed = true
+
+					break
```
JoTurk, line 3804: This can make `writeLoop` block forever. Please propagate the error.

JoTurk (commented):
sorry forgot to send my pending comment

PR #495, opened 2026-09-11
Ignore duplicate INIT on established association

idy:
## What changed

Ignore an INIT received in the Established state when it matches the current association (same port pair and the peer's initiate tag equals our peer verification tag), without mutating the TCB. INIT packets that represent a different association are still handled as before.

## Why

A retransmitted INIT can arrive after its COOKIE ECHO has already moved the receiver to Established. Treating that delayed packet as an unexpected INIT tears down an otherwise healthy association under load.

## Validation

- `go vet ./...`
- `go test -race ./...`


JoTurk (approved):
Good fix for now, but we should implement it properly when we implement assoc restart

PR #500, opened 2026-09-11
Add option to discard inbound data after Close

idy:
## What changed

Add `WithDiscardInboundAfterClose`, an association option (off by default). With it enabled, `Stream.Close` also drops the stream's queued inbound data, and any inbound data that arrives later on that stream is discarded as it arrives. The chunks are still acknowledged, so the sender does not retransmit them. A SACK is sent right away to advertise the reopened receive window. `Read` keeps blocking until the peer resets its outgoing stream and then returns `io.EOF`, as before.

Without the option, behavior is unchanged: `Close` only closes the write direction and inbound data stays readable.

## Why

`Stream.Close` only closes the write direction, so data the peer keeps sending on that stream is queued for `Read` until the peer resets its side. If the application never reads it, as with a WebRTC data channel that is closed without being drained, that data keeps occupying the association's receive window. The peer only resets its outgoing stream after sending all the data it has queued, and sending that data needs the same window. With a large enough backlog, every stream on the association stalls.

The WebRTC data channel model already discards messages received once a channel is closing or closed, so applications on that model lose nothing by dropping them at the SCTP layer. The option is opt-in because other users of `Stream` may rely on reading after a half-close.

## Validation

- new `TestAssocDiscardInboundAfterClose`: the peer sends more than the receive buffer on a stream that is closed without being read. With the option, that data is drained, the window reopens, another stream keeps working, and the peer's reset still ends `Read` with `io.EOF`. Without the option, data sent after `Close` is still readable.
- new `TestStreamDiscardInboundAfterAssociationClosed`: closing after the association is gone still drops the data but does not schedule a SACK.
- option propagation through both the option and the `Config` paths
- `go test -race ./...`, golangci-lint, `GOOS=js GOARCH=wasm go vet ./...`


sirzooro:
@idy please also check https://github.com/pion/sctp/pull/245 . I use that solution for some time without issues. It was merged to pion and then reverted because some test failed due to it. Option to explicitly enable it and keep existing behavior by default would help with that test.

PR #504, opened 2026-09-15
Allow maxRetransmits retransmissions

a-pechenyi:
## Description

A stream with `ReliabilityTypeRexmit` and a reliability value of N gets
N-1 retransmissions instead of N. With `maxRetransmits: 1` a chunk is
abandoned at its first transmission and is never retransmitted, so the
channel behaves exactly like `maxRetransmits: 0`.

`pion/webrtc` passes `RTCDataChannelInit.maxRetransmits` straight through
as the DCEP Reliability Parameter, and `pion/datachannel` passes that to
`Stream.SetReliabilityParams(…, ReliabilityTypeRexmit, value)`, so this
affects every partially reliable DataChannel with `maxRetransmits >= 1`,
whichever side opened it.

### Cause

`checkPartialReliabilityStatus` is called right after `nSent` has been
incremented for the transmission being made:

| Call site | `nSent` at the check |
| --- | --- |
| `movePendingDataChunkToInflightQueue` (first send) | `nSent = 1` |
| `gatherOutboundFastRetransmissionPackets` | `nSent++` |
| `getDataPacketsToRetransmit` (T3-rtx) | `nSent++` |

So `nSent` counts transmissions, including the first one. The check is

```go
if chunkPayload.nSent >= stream.reliabilityValue {
	chunkPayload.setAbandoned(true)
```

which abandons a chunk once its transmissions (not retransmissions) reach
the limit:

| reliability value | transmissions today | retransmissions today | expected retransmissions |
| --- | --- | --- | --- |
| 0 | 1 | 0 | 0 |
| 1 | 1 | **0** | 1 |
| 2 | 2 | **1** | 2 |
| N | N | **N-1** | N |

The chunk that is marked abandoned is still put on the wire this time, so
the last line of the table is exact: the abandonment one transmission too
early costs exactly one retransmission.

The original forward TSN implementation (5094c9c) compared
`d.nSent > s.reliabilityValue` before incrementing `nSent`, which was
correct. b908663 ("Add partial reliability tests and fixed some bugs",
2019) moved `d.nSent++` in front of the check and changed `>` to `>=` in
the same hunk, and it has been `>=` since.

### What the specifications say

- [RFC 7496 Section 3.1](https://www.rfc-editor.org/rfc/rfc7496#section-3.1),
  Limited Retransmissions Policy:

  > The sender MUST abandon a user message if the number of
  > retransmissions of any of the DATA chunks of the user message would
  > exceed the provided limit. [...] Please note that the number of
  > retransmissions includes both fast and timer-based retransmissions.
  >
  > The sender MAY limit the number of retransmissions to 0. This will
  > result in abandoning the message when it would get retransmitted for
  > the first time.

  With a limit of 1, the first retransmission does not exceed the limit.

- [RFC 8832 Section 5.1](https://www.rfc-editor.org/rfc/rfc8832#section-5.1),
  `DATA_CHANNEL_PARTIAL_RELIABLE_REXMIT` and `…_UNORDERED`:

  > User messages will not be retransmitted more times than specified in
  > the Reliability Parameter.

- [RFC 8831 Section 6.1](https://www.rfc-editor.org/rfc/rfc8831#section-6.1)
  makes the RFC 7496 limited retransmission policy mandatory for data
  channels.

- [W3C WebRTC, `RTCDataChannelInit.maxRetransmits`](https://www.w3.org/TR/webrtc/#dom-rtcdatachannelinit-maxretransmits):

  > Limits the number of times a channel will retransmit data if not
  > successfully delivered.

### What other implementations do

usrsctp and dcSCTP both allow N retransmissions:

- usrsctp sets `snd_count = 1` on the first transmission
  ([`sctp_output.c`](https://github.com/sctplab/usrsctp/blob/fd070e05a7474f38c7fecdf4d4b6005d2547ee00/usrsctplib/netinet/sctp_output.c#L7683))
  and drops a chunk only when `snd_count > limit`, both on fast
  retransmit
  ([`sctp_indata.c`](https://github.com/sctplab/usrsctp/blob/fd070e05a7474f38c7fecdf4d4b6005d2547ee00/usrsctplib/netinet/sctp_indata.c#L3611))
  and on T3-rtx
  ([`sctp_timer.c`](https://github.com/sctplab/usrsctp/blob/fd070e05a7474f38c7fecdf4d4b6005d2547ee00/usrsctplib/netinet/sctp_timer.c#L643)).
- dcSCTP retransmits a nacked chunk while
  `num_retransmissions_ < max_retransmissions_` and abandons it otherwise
  ([`outstanding_data.cc`, `Item::Nack`](https://webrtc.googlesource.com/src/+/refs/heads/main/net/dcsctp/tx/outstanding_data.cc)).

## Change

`nSent >= reliabilityValue` becomes `nSent > reliabilityValue`, with a
comment saying what `nSent` counts.

- A reliability value of 0 is unchanged: `nSent = 1 > 0`, so the chunk is
  still abandoned at its first transmission. The existing
  `TestAssocUnreliable` "Rexmit" subtests, which all use 0, pass
  unmodified.
- For N >= 1 the chunk now gets N retransmissions, fast and T3-rtx
  combined, and is abandoned when the N-th is sent.
- `ReliabilityTypeTimed` is not touched.

## Testing

`TestAssocUnreliableRexmitLimit` writes one message on an unordered
stream, drops the first K copies of its TSN at the bridge, and counts how
many copies go on the wire and whether the message is delivered. A few
small messages written after it produce the gap reports; the RTO is
pinned at 50 ms so T3-rtx retransmissions land inside a 1 s window, which
is also long enough for one extra retransmission to show up if the limit
were exceeded.

| limit | copies dropped | expected transmissions | expected delivered |
| --- | --- | --- | --- |
| 0 | 1 | 1 | no |
| 1 | 1 | 2 | yes |
| 1 | 2 | 2 | no |
| 2 | 2 | 3 | yes |
| 2 | 3 | 3 | no |

On `main` (c1b597c) four of the five cases fail:

```
--- FAIL: TestAssocUnreliableRexmitLimit (5.62s)
    --- PASS: TestAssocUnreliableRexmitLimit/limit_0_drops_1 (1.12s)
    --- FAIL: TestAssocUnreliableRexmitLimit/limit_1_drops_1 (1.12s)   transmissions 1, want 2; not delivered
    --- FAIL: TestAssocUnreliableRexmitLimit/limit_1_drops_2 (1.13s)   transmissions 1, want 2
    --- FAIL: TestAssocUnreliableRexmitLimit/limit_2_drops_2 (1.12s)   transmissions 2, want 3; not delivered
    --- FAIL: TestAssocUnreliableRexmitLimit/limit_2_drops_3 (1.13s)   transmissions 2, want 3
```

With the change all five pass.

- `go test ./...`
- `go test -race -count=5 -run TestAssocUnreliableRexmitLimit .`
- `golangci-lint run ./...` (v2.10.1, as in CI): 0 issues


JoTurk (approved):
Thank you for the fix.

Can you also please check https://github.com/pion/webrtc/wiki/Contributing#ai-assisted-pull-requests

PR #511, opened 2026-09-16
Notify when a stream reset completes

idy:
Applications that reuse stream IDs (WebRTC data channels) cannot tell when a reset has finished, and reusing an ID too early lets late reset traffic for the old stream hit the new one. This adds OnStreamResetComplete, called once after both the incoming and outgoing resets are done and the stream is removed.

Split out of #494; overlaps with #463.


association_stream_reset_complete_test.go
```diff
@@ -0,0 +1,162 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
```
JoTurk, line 2: do we really need all of these test? :)

idy: Trimmed: the two tests are now one 84-line test over the bridge (was 162 lines). It still covers the three behaviours that matter: no completion after only one direction is reset, one notification per peer once both are done and the stream is gone, and no completion when only one direction of a reused stream ID is closed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

association_stream_reset_complete_test.go
```diff
@@ -0,0 +1,84 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
```
JoTurk: sorry another request can we move this file to association.go, We're going to enforce this soon via lint but we don't allow standalone test files especially in the root directory

